PR #19, opened 2020-07-02
fix: add app label

open-dynaMIX:


open-dynaMIX:
> LGTM, but shouldn't we, for consistency's sake, rename the module to `alexandria.alexandria_core` as well?

:rofl: I wanted to do this for emeis, which led to a discussion about if this is really needed. So I changed it. In order to avoid another discussion, I did it like that now in alexandria.

And here we are :)

PR #22, opened 2020-07-03
feat(files): add file models and api

open-dynaMIX:
This commit adds...

 - ... a `File` model
 - ... its corresponding views
 - ... configs and container for minio
 - ... the minio storage_client


open-dynaMIX:
> Just a quick question: what's with the commented-out code? Leftovers or stuff that's just not finished yet? I'd probably leave it out of the PR entirely and add it when it's needed

Yes, it's stuff we will need soon. I've removed it for now.

PR #24, opened 2020-07-03
fix: categories are read-only on the API

open-dynaMIX:


winged:
Why shouldn't categories be writable? Shouldn't this be defined via permissions / visibilities as everywhere else?

Maybe it's time to have an overarching design document in here so reviewers like me who weren't fully involved in the planning can see where the project is heading

winged:
> @winged See the RFC here: [projectcaluma/caluma#1105](https://github.com/projectcaluma/caluma/issues/1105)
> Categories always come with custom code defining specific permissions. Managing the categories will be done by developers initially, and potentially by administrators in the future. So, for now, we don't need the categories to be writable.

Created an issue (#30 ) to copy over that proposal to the documentation in this repo

PR #40, opened 2020-07-15
docs(rfc): add original rfc

### LINKED ISSUE #30: Add documentation / Design overview
The [originating issue](https://github.com/projectcaluma/caluma/issues/1105) has a nice write-up on what Alexandria is supposed to be. I suggest copying the issue more or less 1:1 into a "docs" folder here and link it from the README

open-dynaMIX:
Closes #30

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,14 @@
 [![Pyup](https://pyup.io/repos/github/projectcaluma/alexandria/shield.svg)](https://pyup.io/account/repos/github/projectcaluma/alexandria/)
 [![Black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/projectcaluma/alexandria)
 
-Document management service
+Our goal is to implement an external document management service to hold and provide uploaded documents.
+Documents can be uploaded and, depending on user access, managed by internal as well as external users.
+
+The goal is NOT to re implement a complex [DMS](https://en.wikipedia.org/wiki/Document_management_system) but rather to have a simple and user-friendly way of managing documents with different permissions.
+
+All User Interface interactions should be as simple as possible and easily understandable.
+
+[Original RFC that led to alexandria](docs/original_alexandria_rfc.md)
 
 ## Getting started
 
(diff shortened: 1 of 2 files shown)


open-dynaMIX:
@winged Warm-up PR ;)

README.md
```diff
@@ -6,6 +6,8 @@
 
 Document management service
```
winged, line 7: I'd add a sentence or two to be more specific - "document management service" is rather generic. Samba for instance is a document management service :)

docs/original_alexandria_rfc.md
```diff
@@ -0,0 +1,66 @@
+## Rationale
+Documents can currently only be added to a form via file question. These do not provide search, tags, thumbnails and some other features needed for a document management service. File questions can only be used inside a form.
+
+## Goal
+Our goal is to implement an external document management service to hold and provide uploaded documents. Ideally the document management service would be independent of caluma.
```
winged, line 5: I think the "goal" section here is a good fragment to copy to the README to give a short introduction

docs/original_alexandria_rfc.md
```diff
@@ -0,0 +1,66 @@
+## Rationale
```
winged, line 1: Why start with H2 headings? `s/^##/#/g` :-)

docs/original_alexandria_rfc.md
```diff
@@ -0,0 +1,66 @@
...
+Documents can currently only be added to a form via file question. These do not provide search, tags, thumbnails and some other features needed for a document management service. File questions can only be used inside a form.
```
winged, line 2: I'd rephrase this slightly to reference caluma. Might not be obvious to "outsiders".

```suggestion
Documents in Caluma can currently only be added to a form via file question. These do not provide search, tags, thumbnails and some other features needed for a document management service. File questions can only be used inside a form.
```

open-dynaMIX:
@winged Thanks for the review! Shouldn't we better consolidate the RFC in the Caluma issue and then copy/paste it here (like we did with emeis)?

BTW: The README is still on cookiecutter level. I will work on that as soon as the configuration part is finished in alexandria.

PR #42, opened 2020-07-15
feat(visibility): add visibilities

open-dynaMIX:


alexandria/settings.py
```diff
@@ -143,7 +153,7 @@ def parse_languages(languages):
     "DEFAULT_AUTHENTICATION_CLASSES": (
         "mozilla_django_oidc.contrib.drf.OIDCAuthentication",
     ),
-    "DEFAULT_PERMISSION_CLASSES": ("rest_framework.permissions.IsAuthenticated",),
+    "DEFAULT_PERMISSION_CLASSES": ("rest_framework.permissions.AllowAny",),
```
open-dynaMIX, line 156: This is just a temporary change until the permissions are implmented.

alexandria/core/apps.py
```diff
@@ -1,6 +1,17 @@
...
 
 
 class DefaultConfig(AppConfig):
     name = "alexandria.core"
     label = "alexandria_core"
+
+    def ready(self):
+        from .models import VisibilityMixin
+
+        # to avoid recursive import error, load extension classes
```
winged, line 13: Comment should be before the import above

alexandria/core/visibilities.py
```diff
@@ -0,0 +1,119 @@
...
+        return queryset
+
+
+class Any(BaseVisibility):
+    """No restrictions, all models are exposed."""
+
+    pass
+
+
+class OwnAndAdmin(BaseVisibility):
```
winged, line 80: Might need a docstring for a quick explanation

alexandria/settings.py
winged: In this case, please add a `# TODO ` comment in the code :)

alexandria/core/visibilities.py
open-dynaMIX: Done

alexandria/settings.py
open-dynaMIX: Done

alexandria/core/apps.py
open-dynaMIX: Changed. Beware: This was an exact copy/paste from caluma and emeis.

PR #45, opened 2020-07-17
feat(thumbnails): implement generation of thumbnails

open-dynaMIX:
This commit...

 - implements a `/hook` endpoint for minio notification
 - implements generation and management of thumbnails for uploaded files
 - adds `family` to `File` for grouping Files
 - adds an `mc` service to docker-compose.yml
       This service will create the bucket and setup the events


alexandria/core/models.py
```diff
@@ -119,6 +121,16 @@ class Document(UUIDModel):
...
+    ORIGINAL = "original"
+    THUMBNAIL = "thumbnail"
+
+    TYPE_CHOICES = (
+        ORIGINAL,
+        THUMBNAIL,
+    )
+    TYPE_CHOICES_TUPLE = ((type_choice, type_choice) for type_choice in TYPE_CHOICES)
+    type = models.CharField(choices=TYPE_CHOICES_TUPLE, max_length=23, default=ORIGINAL)
+    family = models.ForeignKey("self", on_delete=models.CASCADE, null=True)
```
winged, line 133: Wouldn't it be cleaner to have `family` named `original` and set `related_name="thumbnails"`? Or what is the family going to be used for?

alexandria/core/tests/test_views.py
```diff
@@ -0,0 +1,143 @@
...
+    HTTP_403_FORBIDDEN,
+)
+
+from alexandria.core.models import File
+
+
+@pytest.fixture
+def remove_temp_files(settings):
+    yield
+    for i in settings.THUMBNAIL_CACHE_DIR.glob("**/*"):  # pragma: no cover
```
winged, line 16: Wouldn't it be cleaner to have a `thumbnail_cache` fixture that uses `tempdir` and mocks `settings.THUMBNAIL_CACHE_DIR` for each run?

alexandria/core/models.py
czosel: I'd agree, I prefer `family` to the case where we're dealing with trees.

alexandria/core/tests/test_views.py
open-dynaMIX: Absolutely! Done.

alexandria/core/models.py
open-dynaMIX: Done as discussed.

open-dynaMIX:
@winged I've evaluated the idea of using a different bucket for the renderings. IMO this will make things unnecessarily complicated. As soon as we decide to also create other renderings besides thumbnails (e.g. doc --> pdf), we may want to create thumbnails for them (or other renderings).

So I'd say we just put everything into the same bucket and accept the fact that we have to sort out requests for files we don't need to process.

winged:
> @winged I've evaluated the idea of using a different bucket for the renderings. IMO this will make things unnecessarily complicated. As soon as we decide to also create other renderings besides thumbnails (e.g. doc --> pdf), we may want to create thumbnails for them (or other renderings).
> 
> So I'd say we just put everything into the same bucket and accept the fact that we have to sort out requests for files we don't need to process.

Fair enough! And should be rather easy to migrate if we change our minds in the future.

Rest of the code is reviewed and LGTM!

PR #51, opened 2020-07-23
feat(category): add color field

### LINKED ISSUE #50: Color field on categories
As a visual aid, we want to assign each category a color. Examples of usage in design can be found [here](https://github.com/projectcaluma/alexandria/blob/main/docs/original_alexandria_rfc.md). 
To support this, we need a color field on `Category` which supports HEX color codes. For a first working POC i think it should be enough to just configure the colors via ENV and not via UI.

open-dynaMIX:
closes #50


winged:
Code change looks good, but wouldn't this be a good example for the `meta` field that we're doing in a lot of other places? I think in Ember-Caluma we're already using it to store rendering info as well. This would also allow us to easily extend it if we need more than just the color.

open-dynaMIX:
@winged That's up to @velrest  to decide. There is a `meta` field already, that could be used. AFAIU the idea is to make this a first-class feature. The upside of doing it this way is you get backend-side validation.

velrest:
IMO since the data-format contained in the color field is consistent over instances and does not require instance based flexibility, i would vote against `meta` and keep the model field. Not sure how meta is used in other projects though :shrug:

winged:
As discussed off-line, we'll add the `color` field

PR #67, opened 2020-08-06
feat(tags): add category filter

### LINKED ISSUE #54: Tags endpoint filters
Once a category is selected, we only want to display tags wich actually have content in this category. Also if multiple tags are selected, we only want to display tags which have content in the current category AND content considering the currently selected Tags.
To accomplish this, we need a `category` and a `tags` filter on the endpoint.

open-dynaMIX:
This commit adds a filter `category` to the `tag` view. This filter
makes sure, only tags are returned, that actually have at least one
document in the provided category.

Closes #54

### DIFF
--- a/alexandria/core/filters.py
+++ b/alexandria/core/filters.py
@@ -1,7 +1,7 @@
 import json
 
 from django.contrib.postgres.fields.jsonb import KeyTextTransform
-from django_filters import Filter, FilterSet
+from django_filters import CharFilter, Filter, FilterSet
 from django_filters.constants import EMPTY_VALUES
 from rest_framework.exceptions import ValidationError
 
@@ -67,3 +67,12 @@ class FileFilterSet(FilterSet):
     class Meta:
         model = models.File
         fields = ["original", "renderings", "type", "meta"]
+
+
+class TagFilterSet(FilterSet):
+    meta = JSONValueFilter(field_name="meta")
+    with_documents_in_category = CharFilter(field_name="documents__category__slug")
+
+    class Meta:
+        model = models.Tag
+        fields = ["with_documents_in_category", "meta"]
--- a/alexandria/core/tests/test_filters.py
+++ b/alexandria/core/tests/test_filters.py
@@ -78,3 +78,20 @@ def test_document_search_filter(
 
     result = resp.json()
     assert len(result["data"]) == amount
+
+
+def test_tag_category_filter(db, document_factory, tag_factory, admin_client):
+    blue = tag_factory(slug="blue")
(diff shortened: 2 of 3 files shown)


velrest:
Is this part from the issue intentionally left out or am i missing something?:
> Also if multiple tags are selected, we only want to display tags which have content in the current category AND content considering the currently selected Tags.

open-dynaMIX:
> Is this part from the issue intentionally left out or am i missing something?:
> 
> > Also if multiple tags are selected, we only want to display tags which have content in the current category AND content considering the currently selected Tags.

As discussed in person: We're not implementing this for the time being.

alexandria/core/filters.py
```diff
@@ -67,3 +67,12 @@ class FileFilterSet(FilterSet):
...
         fields = ["original", "renderings", "type", "meta"]
+
+
+class TagFilterSet(FilterSet):
+    meta = JSONValueFilter(field_name="meta")
+    category = CharFilter(field_name="documents__category__slug")
+
+    class Meta:
+        model = models.Tag
+        fields = ["meta"]
```
winged, line 78: Should contain "category" as well, no?

alexandria/core/filters.py
```diff
@@ -67,3 +67,12 @@ class FileFilterSet(FilterSet):
     class Meta:
         model = models.File
         fields = ["original", "renderings", "type", "meta"]
+
+
+class TagFilterSet(FilterSet):
+    meta = JSONValueFilter(field_name="meta")
+    category = CharFilter(field_name="documents__category__slug")
```
winged, line 74: I'm not sure about the intention behind this filter: You're searching for tags with documents in the given category? What's the use case here?

alexandria/core/filters.py
open-dynaMIX: Done!

alexandria/core/filters.py
open-dynaMIX: Use case: A user has selected a category and wants to filter the documents in this category based on tags. For this, we only want to present the tags that have at least one document tagged with it in this category.

I have to admint, that the name `category` for this filter is not very descriptive. An alternative could be `documents-category-slug`, but that's rather vebose. Do you have something in mind?

winged: Aah okay that makes sense. Maybe something along the line of `with-documents-in-category`? I don't really mind the verbosity tbh, we're not on DOS anymore with 8 char names :stuck_out_tongue:

open-dynaMIX: I went with `with-documents-in-category`.

PR #68, opened 2020-08-06
feat(document): add search filter

open-dynaMIX:


alexandria/core/views.py
```diff
@@ -52,6 +52,11 @@ class DocumentViewSet(PermissionViewMixin, VisibilityViewMixin, views.ModelViewS
     serializer_class = serializers.DocumentSerializer
     queryset = models.Document.objects.all()
     filterset_class = DocumentFilterSet
+    search_fields = (
+        "title",
```
winged, line 56: Consider using [`@` prefix on the fields](https://www.django-rest-framework.org/api-guide/filtering/#searchfilter) to enable PostgreSQL FTS search mode

open-dynaMIX: I prefer having the default case-insensitive partial matches here.

With postgres' full-text search, `filename` doesn't match `filename.pdf` for example.

PR #96, opened 2020-09-29
feat(example data): add a command to load example data

velrest:


open-dynaMIX (changes requested):
Nice!

`initial_data.json` only needs to contain records for the models of our own app (`alexandria_core`). So you can remove everything except the categories at the bottom.

Edit: Also a small test would be nice to confirm, that loading the data actually works.

velrest:
@open-dynaMIX added a tiny test

alexandria/core/tests/test_example_data.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def test_load_example_data(db):
+
+    assert len(Category.objects.all()) == 0
+
+    call_command("loaddata", "initial_data.json")
+
+    assert len(Category.objects.all()) == 5
+    assert Category.objects.first().color == "#CB68C1"
```
open-dynaMIX, line 15:
```suggestion
    assert Category.objects.get(pk="alle-beteiligten").color == "#CB68C1"
```

open-dynaMIX (changes requested):
Awesome!

Just a small nitpick.

initial_data.json
```diff
@@ -0,0 +1,92 @@
...
+            "created_by_user": null,
+            "created_by_group": null,
+            "modified_at": "2020-08-04T13:24:00.704Z",
+            "modified_by_user": null,
+            "modified_by_group": null,
+            "meta": {
+
+            },
+            "name": "{\"en\": \"Alle Beteiligten\", \"de\": \"Alle Beteiligten\"}",
+            "description": "{\"en\": \"\"}",
```
open-dynaMIX, line 16: As this is our demo data, please be consistent with translations. Most objects here have a name in german and a description in english.

open-dynaMIX (changes requested):

initial_data.json
velrest: I've added Englisch names everywhere

open-dynaMIX (approved):

PR #110, opened 2020-11-03
fix(slug models): make creation of slug models work correctly

winged:
When creating slug model objects, the slug (primary key) was not generated
(and also not accepted via API, as it was the primary key), thus we would create
an object with an empty-string primary key, which obviously is not going to work

cd @fkm-adfinis

### DIFF
--- a/alexandria/core/serializers.py
+++ b/alexandria/core/serializers.py
@@ -1,4 +1,6 @@
 from django.contrib.auth.models import AnonymousUser
+from django.template.defaultfilters import slugify
+from django.utils import translation
 from rest_framework.exceptions import ValidationError
 from rest_framework_json_api import serializers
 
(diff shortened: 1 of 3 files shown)


alexandria/core/serializers.py
```diff
@@ -37,16 +39,39 @@ class Meta:
         )
 
 
-class CategorySerializer(BaseSerializer):
+class SlugModelSerializer(BaseSerializer):
+    """
+    Ensure on creation that the model will recevie a slug.
```
sbor23, line 44:
```suggestion
    Ensure on creation that the model will receive a slug.
```

sbor23: Funny that gh thinks "Ensure" is a class...

alexandria/core/serializers.py
```diff
@@ -37,16 +39,39 @@ class Meta:
...
+
+    If no slug is passed via API, one is derived from the `name` field of
+    the model, via Django's `slugify()` helper. If the model in question
+    does not have a `name` field, you can specify a `slug_source_field`
+    attribute on the serializer subclass responsible to tell it which
+    field to use instead.
+    """
+
+    def create(self, validated_data):
+        slug_field = getattr(self, "slug_source_field", "name")
```
sbor23, line 54: I suggest the "slug_source_field" to live on the Meta class.

alexandria/core/tests/test_slugmodels.py
```diff
@@ -0,0 +1,39 @@
...
+
+import pytest
+from django.urls import reverse
+
+
+@pytest.mark.freeze_time("2017-05-21")
+def test_create_tags(admin_client):
+    url = reverse("tag-list")
+    # create multiple to ensure PK generation works
+    for name in ["foo", "bar", "baz"]:
```
sbor23, line 11: This would do nicely as pytest parameters. :)

sbor23 (changes requested):

alexandria/core/tests/test_slugmodels.py
winged: Not really. The reason we didn't discover it before was that we only created a single tag, which received the slug "" (empty string). I explicitly want multiple objects created in this run so this won't slip through.

alexandria/core/serializers.py
winged: Good idea, fixing

alexandria/core/serializers.py
winged: Guess only in the suggested-change view :)

sbor23 (approved):

alexandria/core/tests/test_slugmodels.py
```diff
@@ -0,0 +1,39 @@
+"""Module to test api in a generic way."""
```
open-dynaMIX, line 1: Copy/Paste?

open-dynaMIX (commented):
Oops! Thanks for fixing!

PR #112, opened 2020-11-05
feat(permissions): authenticated user checks

winged:
Implement visibility and permission classes that enforce users to be
authenticated.

Extend documentation and while we're at it, document some other important
options as well.


open-dynaMIX (approved):
:tada:

PR #114, opened 2020-11-18
fix(validation): correctly set created-by-(user/group)

winged:
The attributes `created-by-group` and `created-by-user` were incorrectly
set for updates only, instead of new instances


open-dynaMIX (approved):
Oops! :wink:

PR #116, opened 2020-11-26
fix(filters): tag filter should accept multiple values

winged:
The tag filter on the documents endpoint should accept multiple tags.
This allows the filter to do `?filter[tags]=foo,bar`, which will only
list documents which have both "foo" and "bar" tags.

cc @fkm-adfinis


open-dynaMIX (approved):
Awesome :tada:

PR #117, opened 2020-11-26
feat: make tag name monolingual

winged:
There are several reasons for this change:

* Tags may only make sense in one of several languages in use.  There is
  no guarantee that every tag will have a suitable translation into all
  other languages either.

* Tag creation is in almost all tagging systems an "ad-hoc" sort of
  work. There is no centralized management of tags, where an admin would
  sift though the tags, translate them, etc

* There is no way to guarantee that a tag created by user A in their
  language will be usable by user B, if they use another language.

The way forward with this is to add a "language" field, so tags can
be filtered to only show the ones given in your current language.
However the exact desired behaviour in this regard is not yet clear,
so we're leaving out this field for now.

So at this point in time, tags are a monolingual feature. We'll need
to determine their beheviour once we have a multilingual application.

cc @fkm-adfinis


alexandria/core/serializers.py
```diff
@@ -70,7 +70,7 @@ def create(self, validated_data):
         slug_field = getattr(self.Meta, "slug_source_field", "name")
         if "slug" not in validated_data:
             slug_source_value = validated_data.get(slug_field)
-            if isinstance(slug_source_value, dict):
+            if isinstance(slug_source_value, dict):  # pragma: no cover
```
open-dynaMIX, line 73: I'd prefer a `pragma: todo cover` here.

open-dynaMIX (approved):
:tada:

winged:
@fkm-adfinis please implement the corresponding change in ember-alexandria :)

fkm-adfinis:
:sunglasses: 

I'll update the addon.

PR #120, opened 2020-12-04
fix: correctly handle created_by_group, updated_by_group

winged:
In some cases (updates!), created_by_group was validated against
the current user instead of the existing object's user. So
the whole validation in BaseSerializer was refactored to work
a lot cleaner.

### DIFF
--- a/alexandria/core/serializers.py
+++ b/alexandria/core/serializers.py
@@ -10,30 +10,38 @@ from . import models
 class BaseSerializer(serializers.ModelSerializer):
     created_at = serializers.DateTimeField(read_only=True)
 
+    def is_valid(self, *args, **kwargs):
+        # Prime data so the validators are called (and default values filled
+        # if client didn't pass them.)
+        self.initial_data.setdefault("created_by_group", self._default_group())
+        self.initial_data.setdefault("modified_by_group", self._default_group())
+        return super().is_valid(*args, **kwargs)
+
     def validate(self, *args, **kwargs):
         validated_data = super().validate(*args, **kwargs)
+
+        user = self.context["request"].user
+        validated_data["created_by_user"] = user.username
+        validated_data["modified_by_user"] = user.username
 
         self.Meta.model.check_permissions(self.context["request"])
         if self.instance is not None:
             self.instance.check_object_permissions(self.context["request"])
 
-        user = self.context["request"].user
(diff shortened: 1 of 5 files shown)


alexandria/core/serializers.py
```diff
@@ -10,30 +10,47 @@
...
     def validate_created_by_group(self, value):
-        return self._validate_group(value, "created_by_group")
+        # Created by group can be set on creation, then must remain constant
+        if self.instance:
+            # can't change created_by_group on existing instances
+            return self.instance.created_by_group
+        else:
+            return self._validate_group(value, "created_by_group")
+
+    def validate_modified_by_user(self, value):
```
open-dynaMIX, line 45: What purpose is there that this can be sent by the client?

alexandria/core/tests/test_views.py
```diff
@@ -201,40 +201,57 @@ def test_hook_view(
...
+    del serialized_model["modified_by_group"]
+
+    if not update:
+        # delete model in the DB so we can then create it via API
+        model.delete()
 
     post_data = {
         # Note: model name pluralisation may not always be with an "s" suffix,
         # but for now, this holds true
         "data": {"attributes": serialized_model, "type": f"{model_name}s"}
```
open-dynaMIX, line 236: You could use `model.__class__._meta.verbose_name_plural` in order to always have the same plural name as rest framework json api.

open-dynaMIX (approved):
Awesome!

Just two questions/comments.

alexandria/core/serializers.py
winged: None really. This bit is just refactoring from the previous "big" `validate()` method. But thinking about it again, I suppose this could be a lot simpler - fix coming in a minute.

Do you think we could set those fields as `read_only` in `Meta`?

alexandria/core/tests/test_views.py
winged: Done, good point!

alexandria/core/serializers.py
open-dynaMIX: > Do you think we could set those fields as read_only in Meta?

I was thinking something like that. But I'm not sure if they then get dropped before saving.

winged: Yep that's my understanding as well. Only works for created_at, as it's handled by the model itself

PR #121, opened 2020-12-08
fix: deduplicate tags

winged:
The with_documents_in_category filter can duplicate tags, so we
deduplicate the results upfront.

This is easier here than to apply selectively when the filter is active,
and shouldn't impact performance.

### DIFF
--- a/alexandria/core/tests/test_filters.py
+++ b/alexandria/core/tests/test_filters.py
@@ -85,18 +85,26 @@ def test_document_search_filter(
 def test_tag_category_filter(db, document_factory, tag_factory, admin_client):
     blue = tag_factory(slug="blue")
     red = tag_factory(slug="red")
-    tag_factory(slug="green")
+    green = tag_factory(slug="green")
     doc1 = document_factory(category__slug="cat1")
     doc1.tags.add(blue)
-    doc2 = document_factory(category__slug="cat2")
-    doc2.tags.add(red)
+    doc2 = document_factory(category_id="cat1")
+    doc2.tags.add(green)
+    doc3 = document_factory(category__slug="cat2")
+    doc3.tags.add(red)
+
+    # one more to test distinctiveness of the result
+    doc4 = document_factory(category_id="cat1")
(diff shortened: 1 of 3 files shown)


fkm (approved):

alexandria/core/tests/test_filters.py
```diff
@@ -85,17 +85,23 @@ def test_document_search_filter(
...
+    doc4 = document_factory(category_id="cat1")
+    doc4.tags.add(blue)
 
     url = reverse("tag-list")
     resp = admin_client.get(url, {"filter[with-documents-in-category]": "cat1"})
     assert resp.status_code == HTTP_200_OK
     result = resp.json()
-    assert len(result["data"]) == 1
+    assert len(result["data"]) == 2
     assert result["data"][0]["id"] == str(blue.pk)
```
open-dynaMIX, line 105: Can we be sure that the order of the tags in the response is consistent? If not, we should do something like:

```python
assert sorted([t["id"] for t in result["data"]]) == [blue.pk, green.pk]
```

open-dynaMIX (changes requested):
Awesome!

Just a small nitpick.

alexandria/core/tests/test_filters.py
```diff
@@ -85,18 +85,26 @@ def test_document_search_filter(
...
     url = reverse("tag-list")
     resp = admin_client.get(url, {"filter[with-documents-in-category]": "cat1"})
     assert resp.status_code == HTTP_200_OK
     result = resp.json()
-    assert len(result["data"]) == 1
-    assert result["data"][0]["id"] == str(blue.pk)
+    assert len(result["data"]) == 2
+
+    returned_tags = [tag["id"] for tag in result["data"]]
+    assert sorted(returned_tags) == sorted(["blue", "green"])
```
open-dynaMIX, line 107: No need to sort `["blue", "green"]` ;)

open-dynaMIX (approved):

PR #126, opened 2020-12-29
feat: with_documents_meta filter for tags

winged:
This allows limiting display of tags to only those that are
attached to documents with a specific meta value.

For example, when used with Caluma, you could put a caluma
case ID in the document's meta field, and then get only the
tags that are used within the context of a given case.


sbor23 (approved):

winged:
cc: @fkm-adfinis

PR #127, opened 2020-12-30
feat(dev): fake auth backend

winged:
Introduce a fake auth backend that can be used mainly for frontend
development, without the need of setting up a full auth infrastructure.

You can set the following env variables to enable it:
```
DEV_AUTH_BACKEND=true
DEBUG=true
```

Note that `DEBUG=true` is explicitly required, so the fake auth backend
doesn't accidentally leak into production.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,13 @@ A list of configuration options which you need
 * Authorization configurations
   * `VISIBILITY_CLASSES`: Comma-separated list of classes that define visibility for all models
   * `PERMISSION_CLASSES` Comma-separated list of classes that define permissions for all models
+
+For development, you can also set the following environemnt variables to help
+you:
+
+  * `DEV_AUTH_BACKEND`: Set this to "true" to enable a fake auth backend that simulates an authenticated user. Requires `DEBUG` to be set to `True` as well.
+  * `DEBUG`: Set this to true for debugging during development. Never enable this in production, as it **will** leak information to the public if you do.
+
 ## Contributing
 
 Look at our [contributing guidelines](CONTRIBUTING.md) to start with your first contribution.
--- a/alexandria/oidc_auth/authentication.py
+++ b/alexandria/oidc_auth/authentication.py
@@ -3,8 +3,9 @@ import functools
 import hashlib
 
 import requests
+from django.conf import settings
 from django.core.cache import cache
-from django.core.exceptions import SuspiciousOperation
+from django.core.exceptions import ImproperlyConfigured, SuspiciousOperation
 from django.utils.encoding import force_bytes
 from mozilla_django_oidc.auth import OIDCAuthenticationBackend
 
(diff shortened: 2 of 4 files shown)


alexandria/oidc_auth/authentication.py
```diff
@@ -82,3 +83,33 @@ def cached_request(self, method, token, cache_prefix):
             func,
             timeout=self.OIDC_BEARER_TOKEN_REVALIDATION_TIME,
         )
+
+
+class DummyAuthenticationBackend(AlexandriaAuthenticationBackend):
```
open-dynaMIX, line 88: To be more inclusive, I suggest not using the term `dummy` but instead some alternative term. Maybe `MimicAuthenticationBackend()` could work?

alexandria/oidc_auth/authentication.py
```diff
@@ -82,3 +83,33 @@ def cached_request(self, method, token, cache_prefix):
...
+    def __init__(self, *args, **kwargs):  # pragma: no cover
+        super().__init__(*args, **kwargs)
+        if not settings.DEBUG:
+            raise ImproperlyConfigured(
+                "Dummy Auth Backend can only be used in DEBUG mode!"
+            )
+
+    def get_introspection(self, access_token, id_token, payload):  # pragma: no cover
```
open-dynaMIX, line 96: If I'm not mistaken, this is only called from within `get_userinfo_or_introspection()`. So there is no need to override it.

open-dynaMIX (changes requested):
Nice work! :tada:

alexandria/oidc_auth/authentication.py
winged: That's what it seemed to me as well, but given that it's a public method...

alexandria/oidc_auth/authentication.py
winged: Renamed.

README.md
```diff
@@ -60,6 +60,13 @@ A list of configuration options which you need
 * Authorization configurations
   * `VISIBILITY_CLASSES`: Comma-separated list of classes that define visibility for all models
   * `PERMISSION_CLASSES` Comma-separated list of classes that define permissions for all models
+
+For development, you can also set the following environemnt variables to help
+you:
+
+  * `OIDC_AUTH_BACKEND`: Set to "alexandria.oidc_auth.authentication.DummyAuthenticationBackend" to enable a dummy auth backend that simulates an authenticated user. Requires `DEBUG` to be set to `True` as well.
```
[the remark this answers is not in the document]
winged: Good point, made it a boolean setting

winged:
Ready for re-review

alexandria/oidc_auth/authentication.py
open-dynaMIX: There are many more instances of the term in this PR.

sbor23: Do you mean "inclusive" in the sense of: Dummies might be offended by this?

winged: @open-dynaMIX right - fixed them all now :)

open-dynaMIX:
You might want to fix the commit message. Otherwise LGTM!

winged:
> You might want to fix the commit message. Otherwise LGTM!

Typo fixed, PR text updated. I'm not gonna rename the branch though! :P

open-dynaMIX:
Sorry for the nagging :innocent:

open-dynaMIX (approved):

PR #128, opened 2020-12-30
Fix tag filter

winged:
The JSONValueFilter gets the possible `meta` lookups from the referenced field, but in a way that didn't work when the referenced field was accessed indirectly (ie. not a field on the current model).

### DIFF
--- a/alexandria/core/filters.py
+++ b/alexandria/core/filters.py
@@ -20,6 +20,10 @@ class JSONValueFilter(Filter):
         except json.decoder.JSONDecodeError:
             raise ValidationError("JSONValueFilter value needs to be json encoded.")
 
+        if isinstance(value, dict):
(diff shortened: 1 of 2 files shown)


sbor23 (approved):
Looks indeed a bit magic but fine with me. good job!

open-dynaMIX:
@winged Oh no, this means we have to fix it everywhere, right? About time to have this in its separate module!

winged:
> @winged Oh no, this means we have to fix it everywhere, right? About time to have this in its separate module!

Yeap, indeed. Do you have a good idea for a name? "django-jsonfield-filter" maybe?

open-dynaMIX:
> Yeap, indeed. Do you have a good idea for a name? "django-jsonfield-filter" maybe?

Sounds good to me!

PR #135, opened 2021-01-26
feat: search document description as well

winged:
When searching, look in the description field also

### DIFF
--- a/alexandria/core/tests/snapshots/snap_test_api.py
+++ b/alexandria/core/tests/snapshots/snap_test_api.py
@@ -40,12 +40,12 @@ snapshots["test_api_list[FileViewSet] 1"] = {
                     "created-at": "2017-05-21T00:00:00Z",
                     "created-by-group": "admin",
                     "created-by-user": "admin",
-                    "download-url": "http://minio/download-url/9336ebf2-5087-d91c-818e-e6e9ec29f8c1_Devon Cooke",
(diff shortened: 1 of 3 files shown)


sbor23:
This is expected to be full-text search, right? Just because localized fields an search can behave unexpectedly

alexandria/core/tests/test_api.py
```diff
@@ -200,7 +200,7 @@ def test_api_patch(fixture, admin_client, viewset, snapshot, minio_mock):
     assert_response(response, context, snapshot, request_payload=data)
 
 
-@pytest.mark.freeze_time("2017-05-21")
+@pytest.mark.freeze_time("2020-05-21")
```
sbor23, line 203: I don't see how this would contribute in solving flaky tests. Anyways something seems terribly off with the snapshots. Did you update the snapshots after you changed the seed?

winged: I didn't really believe it'd fix it either. I did update the snapshots. Still experimenting how to fix in CI - still works on my machine :tm:

winged:
> This is expected to be full-text search, right? Just because localized fields an search can behave unexpectedly

It's not full-blown FTS, just substring match. Also, as it works (and is unit-tested) for the `title` field on the same model, which is also a localized field, I didn't bother testing it explicitly

open-dynaMIX (approved):
:tada:

alexandria/core/tests/test_api.py
winged: I think it was an outdated image. Updating it on my machine solved the issue

PR #157, opened 2021-04-06
feat: custom validations

winged:
Introduce custom validation layer, similar to the already-existing
permission and visibility classes.

This allows implementors to define additional validations and extend Alexandria.
Usesful for example when incoming data needs to be checked for additional criteria.

### DIFF
--- a/alexandria/core/apps.py
+++ b/alexandria/core/apps.py
@@ -11,10 +11,15 @@ class DefaultConfig(AppConfig):
         # to avoid recursive import error, load extension classes
         # only once the app is ready
         from .models import VisibilityMixin, PermissionMixin
+        from .serializers import BaseSerializer
 
         PermissionMixin.permission_classes = [
             import_string(cls) for cls in settings.PERMISSION_CLASSES
         ]
         VisibilityMixin.visibility_classes = [
             import_string(cls) for cls in settings.VISIBILITY_CLASSES
+        ]
+
+        BaseSerializer.validation_classes = [
+            import_string(cls) for cls in settings.VALIDATION_CLASSES
         ]
--- a/alexandria/core/serializers.py
+++ b/alexandria/core/serializers.py
@@ -10,6 +10,20 @@ from . import models
 class BaseSerializer(serializers.ModelSerializer):
     created_at = serializers.DateTimeField(read_only=True)
 
+    validation_classes = []
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self._custom_validators = []
+
+        validators = [cls(self) for cls in self.validation_classes]
+        for validator in validators:
+            for method_name in [m for m in dir(validator) if not m.startswith("_")]:
+                method = getattr(validator, method_name)
+                if getattr(method, "_validate_model", None) == self.Meta.model:
+                    self._custom_validators.append(method)
+
     def is_valid(self, *args, **kwargs):
         # Prime data so the validators are called (and default values filled
         # if client didn't pass them.)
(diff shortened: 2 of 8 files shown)


sbor23 (approved):
Good job! Just a comment, but I think this should be fine.

open-dynaMIX:
Sorry for chiming in from my holidays. Got a notification on my mobile and couldn't resist looking :see_no_evil: 

I was expecting the need for something like that. IMO the way to go would be to integrate this into [django-generic-api-permissions](https://github.com/adfinis-sygroup/django-generic-api-permissions) and then depending on that in Alexandria. When doing so, we should also import the permission and visibility stuff (mixins, etc) from there and remove any dupicate code from Alexandria. It was already the plan to switch to django-generic-api-permissions anyway, I think now is the right time.

So, now I'll try to turn off the notifications. See you in a week :wave:

winged:
> Sorry for chiming in from my holidays. Got a notification on my mobile and couldn't resist looking see_no_evil
> 
> I was expecting the need for something like that. IMO the way to go would be to integrate this into [django-generic-api-permissions](https://github.com/adfinis-sygroup/django-generic-api-permissions) and then depending on that in Alexandria. When doing so, we should also import the permission and visibility stuff (mixins, etc) from there and remove any dupicate code from Alexandria. It was already the plan to switch to django-generic-api-permissions anyway, I think now is the right time.
> 
> So, now I'll try to turn off the notifications. See you in a week wave

Absolutely agree! But given Alexandria doesn't use this yet, I'd prefer having the implementation here first, let it mature a bit, and then migrate it to django-generic-api-permissions.

winged:
I've rewritten the code to use "decorator style" that we used for visibilities and permissions, instead of class-declarative used before. @sbor23 could you do a re-review please :)

alexandria/core/apps.py
```diff
@@ -11,10 +11,15 @@ def ready(self):
...
         PermissionMixin.permission_classes = [
             import_string(cls) for cls in settings.PERMISSION_CLASSES
         ]
         VisibilityMixin.visibility_classes = [
             import_string(cls) for cls in settings.VISIBILITY_CLASSES
         ]
+
+        BaseSerializer.validation_classes = [
+            import_string(cls) for cls in settings.VALIDATION_CLASSES
+        ]
```
sbor23, line 25: Do we still potentially allow multiple classes? Would be consistent to the other extensions...

alexandria/core/serializers.py
```diff
@@ -10,6 +10,20 @@
...
+        super().__init__(*args, **kwargs)
+
+        self._custom_validators = []
+
+        validators = [cls(self) for cls in self.validation_classes]
+        for validator in validators:
+            for method_name in [m for m in dir(validator) if not m.startswith("_")]:
+                method = getattr(validator, method_name)
+                if getattr(method, "_validate_model", None) == self.Meta.model:
+                    self._custom_validators.append(getattr(validator, method_name))
```
sbor23, line 25:
```suggestion
                method = getattr(validator, method_name)
                if getattr(method, "_validate_model", None) == self.Meta.model:
                    self._custom_validators.append(method)
```

alexandria/core/validation.py
```diff
@@ -0,0 +1,51 @@
...
+
+    return decorator
+
+
+class BaseValidator:
+    """Validate incoming data.
+
+    The validation methods you define here are called at the `validate()` stage
+    of the Rest Framework serializer. You can register any method function in your
+    validator.
```
sbor23, line 23: Only method
```suggestion
    validator. Methods starting with underscore `_` will be ignored.
```

sbor23 (changes requested):

alexandria/core/apps.py
winged: Yes, we support multiple classes. What we don't allow is using the same validator method for multiple models

alexandria/core/validation.py
winged: Fix'd

alexandria/core/serializers.py
winged: Good catch, fix'd

sbor23 (approved):
LGTM!

PR #159, opened 2021-04-07
chore: initial release 0.1.0

winged:


sbor23:
Hmm I don't really see the point of having individual commits listed as features and fixes for the initial release.
We could skip that I think, as all the features should be documented somewhere in the README.md or other docs...

winged:
> Hmm I don't really see the point of having individual commits listed as features and fixes for the initial release.
> We could skip that I think, as all the features should be documented somewhere in the README.md or other docs...

Yeah I don't like the aesthetics either, but I'd still keep it for consistency: Every release will have the exact same format, generated by `python-semantic-release`.

This is also only a problem for the first one :)

sbor23 (approved):
Let's leave it like this for (future) consistency's sake

PR #199, opened 2021-07-12
fix(dev-env): set DEV_AUTH_BACKEND to true

czosel:
This way, the dev env works out of the box.


czosel:
Closing in favor of #200

PR #235, opened 2022-03-10
fix(core): delete unused tags

trowik:


alexandria/core/tests/test_views.py
```diff
@@ -302,3 +303,41 @@ def test_multi_download_failure(admin_client, params, expected_status):
...
+                }
+            },
+        }
+    }
+
+    response = admin_client.patch(url, data)
+
+    assert response.status_code == HTTP_200_OK
+    assert Tag.objects.all().count() == 3
+    assert Tag.objects.filter(slug__in=[tag_1.slug, tag_2.slug, tag_3.slug])
```
winged, line 343: I think this last `assert` line here doesn't really help. It will never fail, as you can search for any slug even ones that don't exist. I suggest using something like this instead:

```python
assert set(Tag.objects.all().values_list('slug', flat=True)) == set([tag_1.slug, tag_2.slug, tag_3.slug])
```

alexandria/core/views.py
```diff
@@ -64,6 +64,27 @@ class DocumentViewSet(PermissionViewMixin, VisibilityViewMixin, views.ModelViewS
...
+                # check if the tag to be deleted is used in other documents
+                # if not delete the tag completely
+                if (
+                    not models.Document.objects.filter(tags=tag)
+                    .exclude(id=instance.id)
+                    .exists()
+                ):
+                    models.Tag.objects.get(slug=tag.slug).delete()
+
+        return super().update(request, *args, **kwargs)
```
winged, line 86: This is quite complicated, and runs a lot of SQL for the expected result. You could just (irregardless of the actual change in the document) do this:

```python
Tag.objects.all().filter(documents__pk__isnull=True).delete()
```

winged (changes requested):

alexandria/core/views.py
trowik: It does delete all currently unused tags, and I have not thought about that.
The only issue I see with your solution is that it doesn't delete tags that are going to be removed by the new request. Deleting a tag completely would require two requests. This could potentially lead to some confusion in frontend, as a deleted tag could still be seen for a while.

winged: Right - this wouldn't work correctly right now. 

There are a few options though:
* You could move the code into a `post_delete` signal handler.
* Call `super().update(...)` first, then do the cleanup before returning - This would be my preferred solution

winged (approved):

PR #239, opened 2022-03-28
chore(deps): update preview-generator to version 0.28

trowik:


alexandria/core/thumbs.py
```diff
@@ -45,6 +46,13 @@ def create_thumbnail(file):
     )
     etag = client.put_object(path_to_preview_image, thumb_file.object_name)
 
+    # remove locked file created by PreviewManager.get_jpeg_preview()
+    # introduced with preview-generator version 0.14
+    # https://github.com/algoo/preview-generator/commit/e8c16499ff38c21ab5a65110fc29635520674ec7#
```
anehx, line 51: I'd put this comment into the commit message since it has no value in the code per se but only in the history of the code..

alexandria/core/thumbs.py
```diff
@@ -45,6 +46,10 @@ def create_thumbnail(file):
     )
     etag = client.put_object(path_to_preview_image, thumb_file.object_name)
 
+    for item in os.listdir(settings.THUMBNAIL_CACHE_DIR):
+        if item.endswith(".lock"):
+            os.remove(os.path.join(settings.THUMBNAIL_CACHE_DIR, item))
```
winged, line 51: I wouldn't want this in the "hot" section of the code. This will loop over all the files in the cache dir every time we create a thumbnail. If I understand correctly, this is a cleanup for a mess created by a previous version of `preview-generator`. So IMHO this is something to do manually, via hint in the release notes.

trowik: Yes, the new version of ```preview-generator``` introduced the lock files.
I thought that with the unlinking we do at the end, it would be nice to delete the lock files as well.

winged: OK, but in that case it does indeed make sense do do a cleanup. But it shouldn't be in the hot path as mentioned before - maybe create a management command for that?

trowik: As discussed I refactored the code to use temporary directory instead of the set in ```settings.py``` (```THUMBNAIL_CACHE_DIR```).

winged (approved):
Nice work! I think it's much cleaner this way now.

Issue #252, opened 2022-07-13
File uploads: Keep track of upload status


winged:
Alexandria stores only metadata for each document / file it tracks. The actual upload is done to Minio.

When something goes wrong in the browser or in Minio, we don't know whether a file has been uploaded correctly or not. Therefore, I suggest adding an `upload_status` flag on the `File` model and it's views to track whether an upload has been completed.

The upload status is updated in the following cases:

* When the thumbnail callback comes in from Minio, telling us that we need to thumbnail a given file
* When the thumbnails have been created and successfully uploaded to Minio

On the REST API, the upload status should thus be readonly.

PR #255, opened 2022-07-14
feat: add upload_status field to file models to track upload status

acortino:
#252


alexandria/core/models.py
```diff
@@ -172,6 +172,22 @@ class File(UUIDModel):
...
+    ERROR = "error"
+
+    STATUS_CHOICES = (
+        UNDEFINED,
+        COMPLETED,
+        ERROR,
+    )
+    UPLOAD_STATUS_TUPLE = (
+        (status_choice, status_choice) for status_choice in STATUS_CHOICES
+    )
```
winged, line 186: This will create a generator and won't be usable twice. Better:

```suggestion
    UPLOAD_STATUS_TUPLE = tuple([
        (status_choice, status_choice) for status_choice in STATUS_CHOICES
    ])
```

winged (changes requested):
Nice work! Just the choices definition in the models needs to be fixed. And maybe add a few lines of explanation to the commit messgage :-)

winged:
Nice, code looks good now. However I'd suggest squashing the commits. No need to have a fix commit for something you did "wrong" in the same PR

Issue #133, opened 2021-01-22
Synonyms for tags


winged:
We need the ability to provide a synonym relationship on tags. Some users may use a slightly different vocabulary for tagging than others, leading to either duplicate tags or documents not being found, because the "wrong" tag is selected.

One could then filter for documents tagged with "house" and also receive documents tagged with "building" for example. This would also help with translation (which we removed from tags), by linking tags of different languages (but the same meaning) together. Extending the example above, you could then also filter for a french tag "maison", and still find documents tagged "building" or "house".

## Implementation

Since multiple tags can be synonyms of each other, a simple M:N relationship would be problematic: Lookups would then need to traverse a forrest of tags to figure out whether two tags belong to each other.

I suggest a new entity `TagSynonymGroup`, to which a tag may optionally belong, via a 1:N relationship. Tag lookups in the documents filterset would then try to lookup a tag directly, OR indirectly via the synonym group, thus being transparent to the user.

Optionally, I think it may be a good idea to extend the "tags" endpoint as well to provide a (read-only) list of related tags. This could then help in the frontend, to show the tag searched for that triggered a match, even if the assigned tag would be another synonym.

The synonym endpoint should of course use the same permissions infrastructure as the documents, and tags themselves.

## Follow-Up (Frontend)

Once implemented, the frontend should be extended to possibly show the synonyms of a tag, for example in a mouseover event, or similar.

Additionally, a management UI for the tags would be needed as well, but can be implemented independently.

winged:
Implemented :)

PR #256, opened 2022-07-14
Feat/tags synonyms

acortino:
#133


alexandria/core/tests/test_filters.py
```diff
@@ -169,6 +171,75 @@ def test_tag_filter(
...
+            obj["id"] for obj in result["data"]
+        ]
+
+    for tag in ungrouped:
+        resp = admin_client.get(url, {"filter[tags]": [tag]})
+        assert resp.status_code == HTTP_200_OK
+        result = resp.json()
+        assert [str(doc.id) for doc in Tag.objects.get(slug=tag).documents.all()] == [
+            obj["id"] for obj in result["data"]
+        ]
```
winged, line 240: This is potentially flaky - if the ordering of the returned documents changes, the filter would still be correct, but the test would fail. We can mitigate this by doing the following:

```suggestion
        assert set([str(doc.id) for doc in Tag.objects.get(slug=tag).documents.all()] )== set([
            obj["id"] for obj in result["data"]
        ])
```

There's the same mistake in the block above on line 230

winged (changes requested):
Looking good apart from a minor detail!

Let's also not go into the complexity of the test function - I think it does the thing it's supposed to do, so let's leave it

PR #261, opened 2022-09-29
fix(Dockerfile): use poetry to run uwsgi

velrest:


winged (changes requested):
LGTM, but the data changes look unrelated / accidental. Did you mean to include them in this PR?

czosel (approved):

PR #285, opened 2023-04-21
Upgrade to django 3.2

Yelinz:
This is a breaking change as drf-jsonapi disallows the use of meta and type as fields in serializers.
This means these fields were renamed to metadata and variant.


winged (approved):
LGTM!

PR #288, opened 2023-05-04
feat: prefix config keys

Yelinz:
deprecate PERMISSION_CLASSES, VISIBILITY_CLASSES, VALIDATION_CLASSES new config keys are
ALEXANDRIA_PERMISSION_CLASSES
ALEXANDRIA_VISIBILITY_CLASSES
ALEXANDRIA_VALIDATION_CLASSES

used https://github.com/projectcaluma/emeis/blob/main/emeis/settings.py as a reference

### DIFF
--- a/alexandria/core/apps.py
+++ b/alexandria/core/apps.py
@@ -14,12 +14,12 @@ class DefaultConfig(AppConfig):
         from .serializers import BaseSerializer
 
         PermissionMixin.permission_classes = [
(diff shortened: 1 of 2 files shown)


winged (approved):
LGTM, but we should probably consider moving to https://github.com/adfinis/django-generic-api-permissions/ while we're at it

Yelinz:
Will do in a separate PR

PR #289, opened 2023-05-04
feat: use dgap instead of custom visibilities and permissions

Yelinz:
~~Problem with CategoryViewSet and FileViewSet, as they are missing the PermissionMixin.
Because they are missing the Create and Destroy mixins/ methods.~~

This refactors the old authentication system to use DGAP instead. Streamlining it, as it is very similar to DGAP.


winged (changes requested):
Nice, looking good! However I think you'll need to mark this as breaking, as it will require changes in the deployments

winged (approved):

Yelinz:
Struggled with updating the snapshots, code did not change.

Yelinz:
Will merge it later, please do not just merge this. Thx

PR #291, opened 2023-05-05
feat: introduce GenericForeignKey to Document

Yelinz:
makemigrations added something for the json fields aswell

A POC on how they work with GenericRelation
https://github.com/Yelinz/django-generic-relation-poc/blob/main/tracking/models.py
https://github.com/Yelinz/django-generic-relation-poc/blob/main/tracking/tests.py

As it stands, it seems pretty usefull for many packages which we release to introduce a GenericForeignKey.
Because it allows the package users to easily create a Many-To-One relation within the orm


winged:
Seems like a good idea at first sight, but we cannot solve this generically.

In my opinion, the whole generic foreign key concept is flawed, as it cannot be indexed properly, it provides no referential integrity apart from possible validation in Django ORM, and worst of all: This will only work when the target model's primary key is also a positive integer field. We often use `UUID`s or slugs (`VARCHAR`), so it will only work in a very limited set of situations.

You *could* try changing the `object_id` type to `CharField(255)`, but still I'm not really convinced.

Your example POC also kinda scares me: Adding a person to a location implicitly removes them from the team they were associated before? This feels like a lot of unexpected behaviour that might bite us hard in the future..

TBH, I think we're still better off with the `metainfo` stuff we're doing right now, even if it's API isn't as clean as this would be.

In my opinion if an application needs to have a N:1 relation from a "foreign" package's model, the right way would be to just do a M:N table, set a proper `related_name` and add the correct constraints / validations.

Sorry to be so negative about this, but I just see this giving us more pain than it takes away

Yelinz:
@open-dynaMIX Regarding the limitations and concerns, I will be using a intermediate table instead of generic relations, we can shelve this for now.

PR #294, opened 2023-05-16
fix: make deprecated VISIBILITY_CLASSES env var optional

czosel:


alexandria/settings/alexandria.py
```diff
@@ -64,10 +64,10 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
 
 
 VISIBILITY_CLASSES = env.list(
-    "VISIBILITY_CLASSES", default=default(["alexandria.core.visibilities.Any"])
+    "VISIBILITY_CLASSES", default=default(["alexandria.core.visibilities.Any"], [])
 )
 PERMISSION_CLASSES = env.list(
-    "PERMISSION_CLASSES", default=default(["alexandria.core.permissions.AllowAny"])
+    "PERMISSION_CLASSES", default=default(["alexandria.core.permissions.AllowAny"], [])
```
winged, line 70: We should do this for the prefixed variants as well as long as the old naming is only deprecated

czosel: fair enough, done

winged (approved):

PR #302, opened 2023-05-31
fix: use correct minio error

Yelinz:


alexandria/core/storage_clients.py
```diff
@@ -23,8 +23,11 @@ def download_url(self, object_name):
                 object_name,
                 timedelta(minutes=settings.MINIO_PRESIGNED_TTL_MINUTES),
             )
-        except minio.error.NoSuchBucket:  # pragma: no cover
-            if settings.MINIO_STORAGE_AUTO_CREATE_MEDIA_BUCKET:
+        except minio.error.S3Error as exc:  # pragma: no cover
```
open-dynaMIX, line 26: `# pragma: no cover` :see_no_evil:

open-dynaMIX (approved):
Change looks good. But it might be worth copying the approach from caluma with its [retry decorator](https://github.com/projectcaluma/caluma/blob/main/caluma/caluma_form/storage_clients.py#L18).

alexandria/core/storage_clients.py
Yelinz: Used the decorator implementation

PR #326, opened 2023-08-14
feat: generate thumbnails without s3 hooks

Yelinz:


winged:
How is this supposed to work? The code instantly triggers the `create_thumbnail()` function, before the *actual file* is even uploaded. This could only work if you're uploading the file directly through Alexandria, which at least in the test case doesn't happen. The thumbnail DB entry is created, but it's of no use if no actual thumbnail is generated.

Sorry for being so blatant, but without *uploading* the actual file, you can't possibly make a thumbnail of it..

alexandria/core/views.py
```diff
@@ -139,10 +139,32 @@ def multi(self, request, **kwargs):
...
+        file = self.get_object()
+        if file.variant == models.File.THUMBNAIL:
+            raise ValidationError(
+                _("File is already a thumbnail, cannot generate thumbnail.")
+            )
+        elif (
+            settings.ALEXANDRIA_ENABLE_THUMBNAIL_GENERATION
+            and not settings.ALEXANDRIA_THUMBNAIL_GENERATE_WITHOUT_HOOK
+        ):
+            raise ValidationError(_("Manual thumbnail generation is not enabled."))
```
winged, line 153: Do we really need to forbid manual thumbnail generation?

README.md
```diff
@@ -72,9 +72,14 @@ A list of configuration options which you need
...
   - `ALEXANDRIA_VALIDATION_CLASSES`: Comma-separated list of classes that define [custom validations](docs/validation.md)
-
-For development, you can also set the following environemnt variables to help
-you:
+- Thumbnail configuration (optional)
+  - `ALEXANDRIA_ENABLE_THUMBNAIL_GENERATION`: Set to `false` to disable thumbnail generation
+    - Check the docker-compose file for an example on how to set up generation with s3 hooks
+  - `ALEXANDRIA_THUMBNAIL_WIDTH`: Width of generated thumbnails
+  - `ALEXANDRIA_THUMBNAIL_HEIGHT`: Height of generated thumbnails
+  - `ALEXANDRIA_THUMBNAIL_GENERATE_WITHOUT_HOOK`: Set to `true` to generate thumbnails without s3 hooks
```
winged, line 80: The setting `ALEXANDRIA_THUMBNAIL_GENERATE_WITHOUT_HOOK` is always checked in tandem with `ALEXANDRIA_ENABLE_THUMBNAIL_GENERATION`. I think we could do completely without this setting.

alexandria/core/views.py
```diff
@@ -139,10 +139,32 @@ def multi(self, request, **kwargs):
...
+
+        create_thumbnail(file)
```
winged, line 155: You should also check if the file already has a thumbnail. This is not done in `create_thumbnail()`, so you may end up with multiple thumbnails if the client calls this endpoint repeatedly

winged (changes requested):
Slowly getting there!

alexandria/core/views.py
```diff
@@ -139,6 +139,27 @@ def multi(self, request, **kwargs):
...
+    def generate_thumbnail(self, request, pk=None):
+        if not settings.ALEXANDRIA_ENABLE_THUMBNAIL_GENERATION:
+            raise ValidationError(_("Thumbnail generation is not enabled."))
+
+        file = self.get_object()
+        if file.variant == models.File.THUMBNAIL:
+            raise ValidationError(
+                _("File is already a thumbnail, cannot generate thumbnail.")
+            )
+        elif file.renderings.all().count() > 0:
```
winged, line 152:
```suggestion
        elif file.renderings.all().exists():
```

This generates better/more performant SQL

alexandria/core/views.py
```diff
@@ -139,6 +139,27 @@ def multi(self, request, **kwargs):
...
+            raise ValidationError(
+                _("File already has thumbnail, cannot generate multiple thumbnails.")
+            )
+
+        create_thumbnail(file)
```
winged, line 157: I wonder if we should check the return status of the `create_thumbnail()` call and generate a corresponding HTTP response.

When we're in a webhook from Minio, there's no recorvery so we don't care, but here, it's (presumably) a human-driven interaction where some success/failure state could be interesting.

winged (changes requested):

alexandria/core/views.py
Yelinz: Returning 400 if it didnt succed

alexandria/core/views.py

README.md
Yelinz: Done

alexandria/core/views.py
Yelinz: Removed

winged (approved):
Nice work, LGTM!

PR #365, opened 2023-10-20
feat: return documents from child categories

Yelinz:


Yelinz:
Not sure if using a raw sql would be better to fetch all children.
Something with `WITH RECURSIVE`

alexandria/core/filters.py
```diff
@@ -114,6 +115,39 @@ def filter(self, qs, value):
...
+        filters = Q()
+        for category in value.split(","):
+            if exclude_children:
+                filters |= Q(category_id=category)
+            else:
+                filters |= Q(
+                    category__slug__in=self._get_child_categories(category),
+                )
+
+        return qs.filter(filters)
```
anehx, line 135:
```suggestion
        filters = Q(category_id=category)
        if not exclude_children:
            filters |= Q(category__parent__slug=category)

        return qs.filter(filters)
```

alexandria/core/filters.py
```diff
@@ -114,6 +115,39 @@ def filter(self, qs, value):
...
+        category = models.Category.objects.filter(slug=category_slug).first()
+        if not category:
+            return []
+
+        children = [category_slug]
+
+        for child in category.children.all():
+            children.extend(self._get_child_categories(child.slug))
+
+        return children
```
anehx, line 148: Let's use filtering by `category__parent__slug` then you don't need this.

anehx (changes requested):

alexandria/core/filters.py
Yelinz: Problem would be that grandchild and further nestedcategories wouldnt work. The filter would only return documents from one nesting level

winged: I think if we need to query categories recursively, we should either use something like [MPTT](https://django-mptt.readthedocs.io/en/latest/), or cache the parents of each category in an `ArrayField`. Recursively iterating in a query is a bad idea performance-wise.

Also `@lru_cache` does not do what you think it does: It uses *all* it's arguments as cache key, so even the `self` parameter, which will be different on the next call (and if you move around categories, it's a whole other can of worms, due to no cache invalidation)

Since we didn't have positive-only experience with MPTT, I would suggest adding a `all_parents = ArrayField(CharField())` to the category model, add a bit of code to keep them updated, and the query basically becomes `all_parents__contains=my_category_slug`

anehx: @winged I think you meant `all_children` - right?

anehx: I agree with @winged. However, I think we need to decide whether we want to do this right now or wait until we have a real use-case for this as currently we don't have this case in reality..

alexandria/core/filters.py
Yelinz: Applied

alexandria/core/filters.py
Yelinz: Ill open an issue for this, but for now we could leave it as it is.

anehx (approved):

PR #382, opened 2023-11-09
feat: add mark model

Yelinz:


winged (approved):
Don't quite understand the purpose of this, but LGTM

Yelinz:
The frontend needed "special" tags which were called marks. But having them seperate on a model makes permissions and other things easier.
This now creates the new marks model and then the frontend can switch from abusing tags to marks.

PR #387, opened 2023-11-16
feat(tags): add new filters, fix search

Yelinz:


alexandria/core/filters.py
```diff
@@ -173,8 +173,10 @@ class TagFilterSet(FilterSet):
     metainfo = JSONValueFilter(field_name="metainfo")
     active_group = ActiveGroupFilter()
     with_documents_in_category = CharFilter(field_name="documents__category__slug")
+    with_documents_in_categories = CharInFilter(field_name="documents__category__slug")
```
czosel, line 176: as discussed: let's return tags with documents in the specified category or any of it's children (to simplify the frontend)

Yelinz: Done

czosel (approved):

PR #389, opened 2023-11-17
feat!: change tag primariy key to uuid

Yelinz:
BREAKING CHANGE: this changes the primariy key for Tag to uuid

- [x] Data migration slug to uuid
- [x] Change field data type to uuid
- [x] Rename field to id

Frontend does not need changes to accommodate this change


alexandria/core/migrations/0012_tag_uuid_schema.py
```diff
@@ -0,0 +1,91 @@
...
+                        t.relname = 'alexandria_core_tag' and a.attname = 'slug'
+                    )or (
+                        t.relname = 'alexandria_core_document_tags' and a.attname = 'tag_id'
+                    ))
+
+            loop
+                EXECUTE format('DROP INDEX %I;', r.index_name);
+            end loop;
+        end;
+        $$ language plpgsql;
```
Yelinz, line 50: Doing this PL/pgSQL command was the only way I found to be able to delete the indexes created by django, as they always have a random string in the index name.

winged: I honestly dislike the SQL approach, even though I understand it's probably the fastest (or even only) way. It does run the risk of having the DB run out of sync with what Django thinks is there, primarily the indexes.

Tried to produce a django-only version, sadly I was unable to get it to work.

Just to make sure we don't miss any constraints / indexes, Could you do the following?

* Dump the DB schema after this migration (only schema, no data)
* Remove all migrations and create them anew, without any magic
* Drop the database and migrate with the "clean" one
* Diff the DB schemas
* Add any missing indexes / constraints, ideally using django's naming convention, or document that this has been done and the indexes are indeed clean (comment on this PR is enough for me)

I'm afraid if we don't do this, we may run into some consistency issues further down the road that may be very hard to track down / debug

alexandria/core/tests/test_migration.py
```diff
@@ -0,0 +1,65 @@
...
+def is_valid_uuid(val):
+    try:
+        uuid.UUID(str(val))
+        return True
+    except ValueError:  # pragma: no cover
+        return False
+
+
+@pytest.mark.django_db()
+def test_tag_pk_migration(migrator):
```
winged, line 15: Had to dig a bit, but: In other projects, we built a `post_migrate_to_current_state` fixture that resets the test DB to the newest version after such a test. The `django-test-migrations` package seems to do this for us, so nice find! (CC @open-dynaMIX @fugal-dy, we should totally replace our ugly fixture with this!)

winged (changes requested):
Nice work! I suggest some validation to ensure the SQL migration is correct and complete in regards to constraints/indexes, but other than that LGTM!

alexandria/core/migrations/0012_tag_uuid_schema.py
Yelinz: After doing this the diff is the foreign key index with the django generated id.

If the entire migration is done in the PL/pgSQL part the index could probably be recreated with the same name.
Other choices would be to create temporary tables where all data is copied over and fresh tables are created, so that django has control over it.

```diff
 --
--- Name: alexandria_core_document_tags alexandria_core_docu_tag_id_fk_alexandri; Type: FK CONSTRAINT; Schema: public; Owner: alexandria
+-- Name: alexandria_core_document_tags alexandria_core_docu_tag_id_8dc7fb00_fk_alexandri; Type: FK CONSTRAINT; Schema: public; Owner: alexandria
 --
 
 ALTER TABLE ONLY public.alexandria_core_document_tags
-    ADD CONSTRAINT alexandria_core_docu_tag_id_fk_alexandri FOREIGN KEY (tag_id) REFERENCES public.alexandria_core_tag(id) DEFERRABLE INITIALLY DEFERRED;
+    ADD CONSTRAINT alexandria_core_docu_tag_id_8dc7fb00_fk_alexandri FOREIGN KEY (tag_id) REFERENCES public.alexandria_core_tag(id) DEFERRABLE INITIALLY DEFERRED;
```

[clean_schema.txt](https://github.com/projectcaluma/alexandria/files/13661861/clean_schema.txt)
[migrated_schema.txt](https://github.com/projectcaluma/alexandria/files/13661862/migrated_schema.txt)

winged: Nice! I think it'd be a good idea to name the new constraint exactly the same way as Django would do it, just so Django can find it again in the future (Assuming the suffix is a hash, not purely random).

Yelinz: Seems to be deterministic, I think we can try to delete the name with the the fixed name, instead of the pgsql
https://github.com/django/django/blob/main/django/db/backends/utils.py#L302
https://github.com/django/django/blob/8fa7c2ae88aee7a2b54a745d25ed38152aad2591/django/db/models/indexes.py#L183C24-L183C36
Ill change the index deletion

Yelinz: [migrated_new_schema.txt](https://github.com/projectcaluma/alexandria/files/13662759/migrated_new_schema.txt)
Now there is no more difference to the django created initial schema.

winged (approved):

winged:
Awesome work!

PR #390, opened 2023-11-21
Use Django storage backend to handle file-upload to s3 storage

fugal-dy:
- [x] add django-storages with s3 (boto3-backed)
 - [x] change file model:
   - [x] new FileField: `content` 
   - [x] get rid of attributes related to the `storage-client` method (upload_url, download_url etc)
   - [x] migrate existing data (basically set the file name in a data migration)
 - [x] setup s3 backend configuration for file upload and encryption
   - [x] dynamic storage selection based on `File.encryption_status`
 - [x] add view-action for file-fetching via backend
 - [x] 100% coverage
 - [x] fix all tests 
 - [x] provide a presigned download url for unauthenticated file retrieval

 - [x] manage command for fetching unencrypted files and reupload encrypted
 - [ ] per object encryption based on object name and secret


alexandria/core/models.py
```diff
@@ -1,14 +1,22 @@
...
 from django.utils.translation import gettext_lazy as _
 from localized_fields.fields import LocalizedCharField, LocalizedTextField
 from rest_framework import exceptions
 
-from .storage_clients import client
+
+def file_directory_path(_file, filename):
+    # directory for File uploads to the local filesystem.
+    return "file/files/{0}/{1}{2}".format(
+        _file.document.pk, str(_file.pk)[:8], filename
```
czosel, line 18: Django File field seems to have built-in filename collision safety, so this should be enough:
```suggestion
    return "file/files/{0}/{1}".format(
        _file.document.pk, filename
```

alexandria/core/serializers.py
```diff
@@ -200,6 +200,34 @@ def validate(self, *args, **kwargs):
...
+        if not isinstance(self.initial_data.get("document"), dict):
+            self.initial_data["document"] = {
+                "type": "documents",
+                "id": self.initial_data["document"],
+            }
+
+        if (original := self.initial_data.get("original")) and not isinstance(
+            original, dict
+        ):
+            self.initial_data["original"] = {"type": "files", "id": original}
```
czosel, line 224: Would this get any easier with the help of `FormDataResourceRelatedField`? See https://github.com/inosca/ebau/blob/main/django/camac/relations.py#L5

alexandria/core/models.py
czosel: It tried uploading two files with the same name in our legacy documents module, which runs with similar config. The second file had some random characters appended to its name to resolve the collision, so I think we don’t have to solve this ourselves.

alexandria/core/views.py
```diff
@@ -36,17 +38,18 @@
...
 class PermissionViewMixin:
     def destroy(self, request, *args, **kwargs):
         self.queryset.model.check_permissions(request)
         instance = self.get_object()
         instance.check_object_permissions(request)
-        # we do not call `super()` in order to not fetch the object twice.
-        self.perform_destroy(instance)
+        # ~~we do not call `super()` in order to not fetch the object twice.~~
+        # NOTE: we call `super()` in order to not circumvent the storage
+        # backend.
```
winged, line 51: I don't think we should keep old comments.. Just drop it

```suggestion
        # NOTE: we call `super()` in order to not circumvent the storage
        # backend.
```

winged (changes requested):
Found a few issues in the code, but mainly smaller ones. The big issues however are:

* Did you verify the data uploaded in the old way is still accessible after the update? That needs to work absolutely certainly, otherwise we'll need a migration of sorts.
* I would like to see the virtualenv/poetry stuff split out into it's own merge request, so we can discuss it separately. I'm not 100% convinced it's the right way, but willing to discuss
* And last: The change needs to be marked breaking, as it's entirely different config, and entirely different way of accessing data for the frontend

alexandria/core/models.py
fugal-dy: done

Dockerfile
```diff
@@ -38,6 +38,10 @@ RUN pip install -U poetry
 
 ARG INSTALL_DEV_DEPENDENCIES=false
 COPY pyproject.toml poetry.lock $APP_HOME/
+ENV VIRTUAL_ENV="$HOME/.cache/pypoetry/virtualenvs/venv"
+ENV PATH="$VIRTUAL_ENV/bin:$PATH" 
+RUN python -m venv $VIRTUAL_ENV
```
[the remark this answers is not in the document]
fugal-dy: I understood that that's how you can get poetry to use a specific virtualenv that is not created with a random name. This change is however dropped and moved to its own PR.

alexandria/core/views.py
fugal-dy: done

Dockerfile
```diff
@@ -38,6 +38,10 @@ RUN pip install -U poetry
...
+RUN poetry install
 RUN if [ "$INSTALL_DEV_DEPENDENCIES" = "true" ]; then poetry install; else poetry install --without dev; fi
```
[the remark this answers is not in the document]
fugal-dy: it's gone

winged (changes requested):
Nice work so far, getting there!

I'd like to see the whole encryption thing to be optional and controllable via settings. Right now you can't really have unencrypted files in S3

alexandria/core/models.py
```diff
@@ -178,19 +185,43 @@ class Document(UUIDModel):
...
+        return _file
+
+
 class File(UUIDModel):
-    ORIGINAL = "original"
-    THUMBNAIL = "thumbnail"
+    class EncryptionStatus(models.TextChoices):
+        SSEC_GLOBAL_KEY = "ssec-global", "SSE-C global key encryption (AES256)"
+        # NOTE: per object encryption is not implemented yet
+        SSEC_OBJECT_KEY = "ssec-object", "SSE-C per object encryption (AES256)"
```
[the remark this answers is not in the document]
fugal-dy: With Django's choices classes `None` is declared in the dunder property `__empty__`. I added the `at-rest` bit of your suggestion, tho

alexandria/storages/fields.py
```diff
@@ -0,0 +1,42 @@
...
+                method := settings.ALEXANDRIA_ENCRYPTION_METHOD
+            ) not in instance.EncryptionStatus.values:
+                msg = (
+                    f"ALEXANDRIA_ENCRYPTION_METHOD must be one of "
+                    f"{instance.EncryptionStatus.values}. {method} is not valid"
+                )
+                raise ImproperlyConfigured(msg)
+            if not (storage := SsecGlobalS3Storage()) and isinstance(
+                storage, S3Storage
+            ):
```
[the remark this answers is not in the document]
fugal-dy: :seal:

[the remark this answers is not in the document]
fugal-dy: that anyway did nothing useful. The check should have been with the default backend to avoid another layer of complexity with potentially different backend-types per file.

alexandria/core/views.py
```diff
@@ -173,82 +169,42 @@ def generate_thumbnail(self, request, pk=None):
...
-        temp_dir.cleanup()
-
-        response_statuses.append(HTTP_201_CREATED)
-
-    # Just return the highest status
-    return HttpResponse(status=max(response_statuses))
+    def create(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        if settings.ALEXANDRIA_ENABLE_AT_REST_ENCRYPTION is True:
```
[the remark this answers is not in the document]
fugal-dy: okay. added the "none" value and changed the `__empty__`  assignment to be as explicit as possible.

fugal-dy:
@anehx Please have a look at 176065115b62cb019ddb338e0779cbe88feb4834 where I refactored your recent addition because the hook view was removed or rather replaced with explicit logic within the view.

alexandria/core/views.py
```diff
@@ -168,87 +171,63 @@ def multi(self, request, **kwargs):
...
+            ] = settings.ALEXANDRIA_ENCRYPTION_METHOD
+        obj = serializer.save()
+        if (
+            settings.ALEXANDRIA_ENABLE_THUMBNAIL_GENERATION
+            and obj.variant == models.File.Variant.ORIGINAL
+        ):
+            try:
+                create_thumbnail(obj)
+            except DjangoCoreValidationError as e:
+                raise ValidationError(*e.messages)
```
winged, line 189: Here's a semantic difference between the new version and the old: The old would upload a file and then just be unable to create a thumbnail. Here, we'll reject the whole upload if the thumbnail fails for some reason.

Should we instead just log an error if thumbnail creation fails? Cause otherwise, we'll sure have some files that just can't be uploaded, with potentially nothing the user can do about it

alexandria/core/views.py
```diff
@@ -168,87 +171,63 @@ def multi(self, request, **kwargs):
...
+            try:
+                # coerce the uploaded file to a thumbnail
+                create_thumbnail(obj)
+            except DjangoCoreValidationError as e:
+                # remove the uploaded file on failure to avoid
+                # collsions
+                obj.delete()
+                raise ValidationError(*e.messages)
+            # if the original already had a thumbnail remove that
+            self.queryset.filter(original=obj.original).exclude(pk=obj.pk).delete()
```
winged, line 200: This should probably be moved to the top of the `if` block, no? Otherwise it will delete the just-created thumbnail

alexandria/core/serializers.py
```diff
@@ -147,39 +153,100 @@ class FileSerializer(BaseSerializer):
...
 
-    download_url = serializers.CharField(read_only=True)
-    upload_url = serializers.SerializerMethodField()
-
-    def get_upload_url(self, obj):
-        return obj.upload_url if self.new else ""
+    @staticmethod
+    def make_signature_components(
+        pk: str, request: Request, expires: Optional[int] = None
+    ) -> Tuple[str, int, str]:
```
winged, line 168: I think this method could be moved into a module context, so it's accessible without the DRF componentry

alexandria/core/views.py
```diff
@@ -168,87 +171,63 @@ def multi(self, request, **kwargs):
...
+        headers = self.get_success_headers(serializer.data)
+
+        return Response(serializer.data, status=HTTP_201_CREATED, headers=headers)
+
+    @permission_classes([AllowAny])
+    @action(methods=["get"], detail=True)
+    def download(self, request, pk=None):
+        if token_sig := request.query_params.get("signature"):
+            now = timezone.now()
+            expires = float(request.query_params.get("expires"))
```
winged, line 210: I would make this an `int`, not a `float` (also on URL generation side - no need to be this precise, and it can be a source of errors)

alexandria/core/views.py
```diff
@@ -168,87 +171,63 @@ def multi(self, request, **kwargs):
...
+            #
+            if now.timestamp() > expires:
+                raise PermissionDenied("Download URL expired.")
+            try:
+                assert token_sig == signature
+            except AssertionError:
+                raise PermissionDenied("Invalid signature.")
+            queryset = self.get_queryset()
+            filter_kwargs = {"pk": pk}
+            obj = get_object_or_404(queryset, **filter_kwargs)
```
winged, line 229:
```suggestion
            obj = self.get_object()
```

winged (changes requested):
Nice work! I think there's only some smaller things to cleanup.

One suggestion not noted in the comments: I'd probably extract the whole signing/validation stuff into it's own module, so it can be called from the serializers and views instead of having an (additional) interdependence between the two

alexandria/core/views.py
fugal-dy: I would say that raising the ValidationError at this point `L189` the code will not roll back the previous `serializer.save()` statement, no? Therefore the upload will still be successful for the actual object but no thumbnail will be available. I agree that it might be confusing for the frontend, since there is no success message for the upload but only an error message regarding the thumbnail.

So, no sure what's the right thing to do here.. The main object has been created. But only a thorough success of creation and configured side effects like thumbnail creation would suggest a return value of 2xx. So maybe I could add to the message that the primary object was created but the thumbnail remained somewhere in a pickle..

alexandria/core/views.py
fugal-dy: That's the indended behavior in my opinion. If it fails to thumbnail the thumbnail from the uploaded file we don't want it laying around. it could be anything, no?

fugal-dy: ~~That's the indended behavior in my opinion. If it fails to thumbnail the thumbnail from the uploaded file we don't want it laying around. it could be anything, no?~~

sorry. I thought you were talking about L197. Since the whole `if` is suitable to replace a thumbnail I would not remove the old before the new one was successfully created. The `.exclude(pk=obj.pk)` should prevent deletion of the new thumbnail

alexandria/core/serializers.py
fugal-dy: If I move it to its own module would you also suggest that I make it non-dependent on the `Request` object? Requiring a host parameter? That was actually the reason to add it there, because that's a context where the request is available normally

alexandria/core/views.py
winged: Ahh yes that absolutely makes sense. You want a fully unauthenticated request to work if you have a correctly signed URL. In that case however, I'd suggest going directly to the `File` model to avoid all DRF queryset handling

alexandria/core/serializers.py
winged: Yes I think that'd be a good idea. Might make the code a bit cleaner too

alexandria/core/views.py
winged: That depends on whether [ATOMIC_REQUESTS](https://docs.djangoproject.com/en/4.2/topics/db/transactions/#tying-transactions-to-http-requests) is enabled or not. In any case I think that when we return an error, the front-end should reasonably expect that the whole request did not go through.

If we store the main file, we should return a proper response, independent of the thumbnail. If we decide that thumbnail creation is integral part of the upload, we should wrap the whole thing in a transaction and rollback if thumbnail creation fails. Followup thought: If we save the thumbnail successfully, it should be in the relationships in the output, and if it failed, there won't be a thumbnail relationship.

Generally I think we should not make the thumbnail a critical element. There may be documents that can't be thumbnailed despite being valid. If we're concerned about the main document being messed up, we should do an explicit validation for that and not rely on the thumbnail success

fugal-dy: Fair points. So I'll do just that: if thumbnail creation was successful, include it in the returned object date. I'll just log an error on failing thumbnail creation if that's configured, so they may be picked up if occurring frequently.

fugal-dy: Followup: the thumbnail is no direct relation of the original.. so that won't work. I will however return a 201 if the original was created successfully. The other option would be to implement a one shot document with files upload endpoint where you could add a new (version of a) file to the document list.. but that would require some more specs.

alexandria/core/views.py
```diff
@@ -94,36 +93,41 @@ def update(self, request, *args, **kwargs):
...
+    def get_serializer_context(self):
+        context = super().get_serializer_context()
+        context.update({"view": self, "request": self.request})
+        return context
+
     def _write_zip(self, file_obj, queryset):
         with zipfile.ZipFile(file_obj, "w", zipfile.ZIP_DEFLATED) as zipf:
             seen_names = set()
-            for file in queryset.order_by("-name").iterator():
+            for _file in queryset.order_by("-name").iterator():
```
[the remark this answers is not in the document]
fugal-dy: :ok_hand:

alexandria/core/serializers.py
fugal-dy: done  in fdb5943

alexandria/core/views.py
fugal-dy: done. it's all `int`s now

alexandria/core/views.py
fugal-dy: done

fugal-dy: also in fdb5943

pyproject.toml
```diff
@@ -66,6 +68,9 @@ snapshottest = "0.6.0"
 werkzeug = "3.0.1"
 
 
+[tool.poetry.group.dev-dependencies.dependencies]
```
Yelinz, line 71: Replace the old poetry format on line 42 with this.
Instead of having duplicate groups

README.md
```diff
@@ -82,6 +82,33 @@ A list of configuration options which you need
...
   - `ALEXANDRIA_ENABLE_CHECKSUM`: Set to `false` to disable file checksums. Checksums are calculated after upload to allow later verification (not implemented in Alexandria)
 
+- Storage configuration
+
+  Storage backends are configured globally. The storable object bears information on the encryption status allowing the ORM appropriate handling of the data.
+
+  - `FILE_STORAGE_BACKEND`: Set the backend for file uploads. `django-storages` is available (default: `django.core.files.storage.FileSystemStorage`)
+
+  Encryption:
+  - `ALEXANDRIA_ENABLE_AT_REST_ENCRYPTION`: Set to `true` to enable at-rest encryption of files (does nothing by itself unless `ALEXANDRIA_ENCRYPTRION_METHOD` is set to a supported method)
```
winged, line 92: I think this is dangerous. Enabling encryption should *force* you to configure everything needed. You shouldn't be able to set `ALEXANDRIA_ENABLE_AT_REST_ENCRYPTION=True` and allow unencrypted files to be stored

alexandria/core/management/commands/encrypt_files.py
```diff
@@ -0,0 +1,26 @@
...
+from alexandria.core.models import File
+
+
+class Command(BaseCommand):
+    help = "Swaps plain text file content to encrypted content"
+
+    def add_arguments(self, parser):
+        parser.add_argument("--dry", dest="dry", action="store_true", default=False)
+
+    def handle(self, *args, **options):
```
winged, line 14: This needs some settings validation. If you run this command without encryption properly setup, it will just re-upload the unencrypted files, but set the encryption method to whatever's in the settings.

We should be sure that encryption is enabled and configured correctly before commencing with this task

alexandria/core/presign_urls.py
```diff
@@ -0,0 +1,54 @@
...
+    It tests against the expiry: raises a TimeoutError
+    It tests against signature integrity: raises an AssertionError
+
+    returns True otherwise.
+    """
+    now = timezone.now()
+    host, expires, signature = make_signature_components(pk, hostname, expires, scheme)
+
+    if int(now.timestamp()) > expires:
+        raise TimeoutError()
```
winged, line 48: I don't think `TimeoutError` is intended for this. I expect a timeout error when a process takes too long (for example when downloading and the server didn't respond in a while).

In the end, this should trigger a 400, as the signature expired. Therefore I'd base the exception here off of DRF's ValidationError instead

winged (changes requested):
Nice work! I think after some final cleanup, this is on a good track :+1:

README.md
Yelinz: Added a check https://github.com/projectcaluma/alexandria/blob/e941029fae378f57e5ace86171472b96b5c89956/alexandria/storages/fields.py#L39-L42

alexandria/core/management/commands/encrypt_files.py
Yelinz: Done
https://github.com/projectcaluma/alexandria/blob/e941029fae378f57e5ace86171472b96b5c89956/alexandria/core/management/commands/encrypt_files.py#L15-L24

Will be adding a test for the command, not sure coverage doesnt complain

alexandria/core/presign_urls.py
Yelinz: Swapped for ValidationError, seems like previously it was used to signal the call site what kind of error it is. So the call site can handle it. (Differentiate between assertion and timeout)
Removed the error handling at the callsite

alexandria/core/presign_urls.py
```diff
@@ -0,0 +1,54 @@
...
+    try:
+        assert token_sig == signature
+    except AssertionError:
+        raise
```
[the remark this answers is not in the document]
Yelinz: Changed to `ValidationError`

winged (changes requested):
Any minute now :-)

winged (approved):

winged:
LGTM now. Are you gonna cleanup the commits a bit before merging? Seems like quite a long chain of fixes...

Yelinz:
trying to merge the commits messed with the history, because there are merges in between.
Rebasing to remove those again is very tedious...

PR #392, opened 2023-11-27
feat(dev): make a userspace mountable virtualenv

fugal-dy:
The `.home` directory holds the in-container virtualenv that can be made `rw` and hence edited for debugging purposes.

You can e. g. `ln -s` to the venv location for a shortcut when you need to edit files or set breakpoints etc. E. g. `.venv`

The same way the virtualenv is available for the languageserver (e. g. pyright) for code inspection, navigate the source, auto-importing etc. so you will not need to maintain an extra virtualenv on the host.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,6 +6,7 @@ docker-compose.*
 Dockerfile
 .dockerignore
 .env
+.venv
 .git
 *.pyc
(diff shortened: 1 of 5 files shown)


fugal-dy:
closing. seems too esoteric

PR #398, opened 2023-12-11
feat(document): add a method to clone documents

JohnL17:


alexandria/core/models.py
```diff
@@ -137,6 +137,22 @@ class Document(UUIDModel):
...
+
+        self.pk = None
+        self.save()
+
+        for file in [latest_original, thumbnail]:
+            file.pk = None
+            file.document = self
+            file.save()
+
+        return self
```
czosel, line 154: Let's check if we actually get two files in minio when cloning. If not, we might run into problems when one of them is deleted.

czosel: added code that copies the file in S3.

czosel:
approved by @anehx

PR #400, opened 2023-12-13
fix: add missing permission mixins

Yelinz:


winged (changes requested):
LGTM so far, except the deleting files part. This should be allowed I think - if an application doesn't want this, they can still do a custom validation to disable it

Yelinz:
Allowing delete on files

winged (approved):

PR #403, opened 2023-12-20
fix: update dgap

Yelinz:
Im not sure why there are a lot more queries now and if that is problematic as they are very small queries.


winged (approved):
LGTM!

PR #404, opened 2023-12-28
fix: dont overwrite created by

Yelinz:


alexandria/core/serializers.py
```diff
@@ -22,10 +22,10 @@ def is_valid(self, *args, **kwargs):
     def validate(self, *args, **kwargs):
         validated_data = super().validate(*args, **kwargs)
 
-        user = self.context["request"].user
-        username = getattr(user, settings.ALEXANDRIA_CREATED_BY_USER_PROPERTY)
-        validated_data["created_by_user"] = username
-        validated_data["modified_by_user"] = username
+        if "created_by_user" not in validated_data:
+            validated_data["created_by_user"] = self._default_group()
```
czosel, line 26: wouldn't this still allow changing `created_by_user` by passing it? After creation, the property should not be changeable.

Yelinz: Added all to DRF `read_only_fields`

Yelinz: added validate_created_by_user

czosel (approved):

PR #412, opened 2024-01-08
chore(release): v3.0.0

Yelinz:


anehx:
@czosel Should we maybe wait for #390 as that will be a breaking change as well? What's the progress there?

czosel:
Hi @anehx, good point. #390 will still take a few weeks to complete, so let's do a RC release.

czosel (approved):

czosel:
@Yelinz this probably needs an update with the latest fix in #413 now.

PR #416, opened 2024-01-12
feat(document): update modfied data when creating new file

Yelinz:


alexandria/core/serializers.py
```diff
@@ -197,6 +197,17 @@ def validate(self, *args, **kwargs):
 
         return validated_data
 
+    def create(self, validated_data):
+        created = super().create(validated_data)
+
+        # Update document's modified when a file is created.
+        created.document.modified_at = created.modified_at
```
open-dynaMIX, line 204: This field has `auto_now=True`, so it's not necessary to set it manually.

open-dynaMIX (changes requested):
Nice work :tada: Just a small change needed.

alexandria/core/serializers.py
Yelinz: Removed the line

PR #424, opened 2024-01-22
fix: fix factory import

Yelinz:
importing from test caused problems after it was packaged, because that subdirectory is excluded


fugal-dy (approved):
looks pretty sane to me :+1:

PR #425, opened 2024-01-22
fix: remove modified at update for file creation 

Yelinz:


fugal-dy (approved):
Looks fair. The question remains why this was there in the first place..

Yelinz:
Frontend was displaying modified_at, which didnt update when uploading a new file. Which doesnt feel intuitive for users. Changed what is being displaying in the frontend, therefore this isnt needed anymore. Its a cleanup

PR #427, opened 2024-01-22
fix(file): add missing create permission check

Yelinz:


Yelinz:
Both are just to check if the permission checks are ran, as they have modified the basic viewset methods. Because currently it slipped through that no permissions are evaluated for files.

fugal-dy (approved):
fair

PR #431, opened 2024-01-26
chore: add management command to generate missing thumbnails

czosel:


alexandria/core/management/commands/generate_missing_thumbnails.py
```diff
@@ -0,0 +1,21 @@
...
+
+    @transaction.atomic
+    def handle(self, *args, **options):
+        for file in tqdm(
+            File.objects.filter(variant="original", renderings__isnull=True)
+        ):
+            create_thumbnail(file)
+            if virtual_memory().available < 300000000:
+                print("about to run out of memory, stopping")
+                break
```
Yelinz, line 21: Will it just fill up memory otherwise?

czosel: Yeah, similar to what we saw with the encryption command

PR #436, opened 2024-01-26
feat(security): add ClamAV integration

Yelinz:
This adds django_clamd for AV scanning capabilites with ClamAV.
The scanning can be enabled through the settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,13 @@ A list of configuration options which you need
     The development setup features a minio service, implementing the S3 protocol.
     To use SSE-C in development make sure to generate a certificate for the minio container and set `ALEXANDRIA_S3_VERIFY` to `false`.
 
+- ClamAV
+- `ALEXANDRIA_CLAMD_ENABLED`: Set this to `True` to enable ClamAV (virus scanner).
+- `ALEXANDRIA_CLAMD_SOCKET`: ClamAV socket
+- `ALEXANDRIA_CLAMD_USE_TCP`: Use TCP to connect to ClamAV service
+- `ALEXANDRIA_CLAMD_TCP_SOCKET`: ClamAV service socket
+- `ALEXANDRIA_CLAMD_TCP_ADDR`: ClamAV service address
+
 For development, you can also set the following environemnt variables to help you:
 
 - `ALEXANDRIA_DEV_AUTH_BACKEND`: Set this to "true" to enable a fake auth backend that simulates an authenticated user. Requires `DEBUG` to be set to `True` as well.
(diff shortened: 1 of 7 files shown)


alexandria/settings/alexandria.py
```diff
@@ -160,3 +160,10 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
 
 # Checksums
 ALEXANDRIA_ENABLE_CHECKSUM = env.bool("ALEXANDRIA_ENABLE_CHECKSUM", default=True)
+
+# Clamav service
+CLAMD_SOCKET = env.str("ALEXANDRIA_CLAMD_SOCKET", default="/var/run/clamav/clamd.ctl")
+CLAMD_USE_TCP = env.bool("ALEXANDRIA_CLAMD_USE_TCP", default=True)
+CLAMD_TCP_SOCKET = env.str("ALEXANDRIA_CLAMD_TCP_SOCKET", default=3310)
+CLAMD_TCP_ADDR = env.str("ALEXANDRIA_CLAMD_TCP_ADDR", default="localhost")
+CLAMD_ENABLED = env.bool("ALEXANDRIA_CLAMD_ENABLED", default=True)
```
winged, line 169:
```suggestion
CLAMD_ENABLED = env.bool("ALEXANDRIA_CLAMD_ENABLED", default=default(False,True))
```
This would disable it in dev, but enable in prod, unless explicitly configured.

I'm also not sure I'd enable it by default at all - no strong opinions, but I feel like this shouldn't be enabled by default

winged (changes requested):
Small nitpick about config defaults.

Also, for new features, I'd like to see a short description in the commit message and pull request to explain a bit more what it does. Generally I'd like to see less one-liners (this is not a personal feedback, just a general observation)

alexandria/settings/alexandria.py
Yelinz: Makes sense otherwise this is breaking.
Changed default to false.

README.md
```diff
@@ -111,10 +111,17 @@ A list of configuration options which you need
...
+- `ALEXANDRIA_CLAMD_SOCKET`: ClamAV socket
+- `ALEXANDRIA_CLAMD_USE_TCP`: Use TCP to connect to ClamAV service
+- `ALEXANDRIA_CLAMD_TCP_SOCKET`: ClamAV service socket
+- `ALEXANDRIA_CLAMD_TCP_ADDR`: ClamAV service address
+
 For development, you can also set the following environemnt variables to help you:
 
 - `ALEXANDRIA_DEV_AUTH_BACKEND`: Set this to "true" to enable a fake auth backend that simulates an authenticated user. Requires `DEBUG` to be set to `True` as well.
 - `DEBUG`: Set this to true for debugging during development. Never enable this in production, as it **will** leak information to the public if you do.
+- `CLAMD_ENABLED`: Set this to `False` to disable clamd scanning in development.
```
winged, line 124:
```suggestion
- `CLAMD_ENABLED`: Set this to `True` to enable ClamAV (virus scanner).
```

winged (changes requested):
Docs should match code. And it's a pity the `pyproject.toml` got shuffled around, even though no actual change happend there despite the new dependency. Especially the coverage config which is entirely unrelated to this change.

README.md
Yelinz: Fixed docs, reverted coverage config

winged: Awesome, nice work!

winged (approved):

PR #437, opened 2024-01-30
feat(storage): don't use django default storage configuration

anehx:
BREAKING CHANGE: Instead of overwriting `DEFAULT_FILE_STORAGE` in the host app, Alexandria now uses a separate setting `ALEXANDRIA_FILE_STORAGE` to configure the used file storage backend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ A list of configuration options which you need
 
   Storage backends are configured globally. The storable object bears information on the encryption status allowing the ORM appropriate handling of the data.
 
-  - `FILE_STORAGE_BACKEND`: Set the backend for file uploads. `django-storages` is available (default: `django.core.files.storage.FileSystemStorage`)
+  - `ALEXANDRIA_FILE_STORAGE`: Set the backend for file uploads. `django-storages` is available (default: `alexandria.storages.backends.s3.S3Storage`)
 
   Encryption:
 
@@ -94,7 +94,7 @@ A list of configuration options which you need
   - `ALEXANDRIA_ENCRYPTION_METHOD`: Define encryption method that is applied to uploaded objects. Available values depend on storage backend's capabilities (default: `None`)
     - available methods
       - None: no at-rest encryption
-      - `ssec-global`: encrypt all files with the same key (requires: `FILE_STORAGE_BACKEND`: `alexandria.storages.s3.S3Storage)
+      - `ssec-global`: encrypt all files with the same key (requires: `ALEXANDRIA_FILE_STORAGE`: `alexandria.storages.backends.s3.S3Storage`)
 
   Supported backends:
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -103,13 +103,13 @@ A list of configuration options which you need
...
     required configuations:
 
-    - `AWS_S3_ACCESS_KEY_ID`: identity
-    - `AWS_S3_SECRET_ACCESS_KEY`: password to authorize identity
-    - `AWS_S3_ENDPOINT_URL`: the url of the service
-    - `AWS_STORAGE_BUCKET_NAME`: the bucket name of the storage to access objects in path notation (not subdomain)
+    - `ALEXANDRIA_S3_ACCESS_KEY`: identity
+    - `ALEXANDRIA_S3_SECRET_KEY`: password to authorize identity
+    - `ALEXANDRIA_S3_ENDPOINT_URL`: the url of the service
+    - `ALEXANDRIA_S3_BUCKET_NAME`: the bucket name of the storage to access objects in path notation (not subdomain)
```
fugal-dy, line 109: I think you cannot change those variable names starting with `AWS_S3_` at will because the s3 storages backend is configured by those in the dependency code

alexandria/core/management/commands/encrypt_files.py
```diff
@@ -36,15 +36,15 @@ def handle(self, *args, **options):
...
+        Storage = get_storage_class(settings.ALEXANDRIA_FILE_STORAGE)
         for file in tqdm(
             File.objects.filter(
                 Q(encryption_status=File.EncryptionStatus.NOT_ENCRYPTED)
                 | Q(encryption_status__isnull=True)
             ),
         ):
             # get original file content
-            file.content.storage = DefaultStorage()
+            file.content.storage = Storage()
```
fugal-dy, line 47: with renaming `DefaultStorage` to `Storage` the information that it's indeed the default storage is lost but that difference is kind of the point here IMO

README.md
fugal-dy: And thinking of that: since those env vars are dependent on which backend is used it would actually make sense to move those to `alexandria.storages.backend.<some_backend>` and load them only when said backend is selected so they would not clutter the alexandria settings proper.

fugal-dy (changes requested):
Fair point with singling alexandria's storage config out.

I would suggest to leave the other varable renaming aside in this PR

alexandria/core/management/commands/encrypt_files.py
anehx: But it's not the default storage anymore but the storage class that is defined in `ALEXANDRIA_FILE_STORAGE` which is the entire point of this PR

README.md
anehx: > I think you cannot change those variable names starting with AWS_S3_ at will because the s3 storages backend is configured by those in the dependency code

Yes, we can because we explicitly read those settings in the storage class: https://github.com/projectcaluma/alexandria/blob/bb8ad5a16beee65212c944b25777aeea0e30b35a/alexandria/storages/backends/s3.py#L7-L19

> And thinking of that: since those env vars are dependent on which backend is used it would actually make sense to move those to alexandria.storages.backend.<some_backend> and load them only when said backend is selected so they would not clutter the alexandria settings proper.

Not sure it's worth the effort to properly load env vars from the storage class file as we'd have problems if this is used as a django app (proper handing of `.env` files etc. which is dependent on the host app). But what we could do is configure the whole storage in an object kind of like the settings in rest framework. Something like:

```python
ALEXANDRIA_STORAGE = {
    "BACKEND": env.str("ALEXANDRIA_STORAGE_BACKEND", default="alexandria.storage.backends.s3.S3Storage")
    "S3_OPTIONS": {
        "BUCKET_NAME":  "..."
    }
}
```

fugal-dy: Haha, looks like I failed to reexamine the diff of the diff etc. after your force push from 7390da3 to 3905881

so apparently we agree.

fugal-dy (approved):

PR #442, opened 2024-02-02
feat(helm): add chart

hairmare:


deploy/charts/alexandria/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
...
+  labels:
+    {{- include "alexandria.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": test
+spec:
+  containers:
+    - name: wget
+      image: busybox
+      command: ['wget']
+      args: ['{{ include "alexandria.fullname" . }}:{{ .Values.service.port }}']
```
hairmare, line 14: it should put a file to alexandria and delete it again

hairmare:
image signing and heml chart pushing might be a tad sketchy because i haven't been able to test it

it happens late in the pipeline after important things already is done and fixing it should be trivial given it's for sure not far off from working

PR #450, opened 2024-02-08
fix(document): fix clone method on document

anehx:
- Actually copy the file on the storage when cloning
- Generate thumbnails for cloned files


alexandria/core/models.py
```diff
@@ -144,17 +151,34 @@ class Document(UUIDModel):
...
+        from alexandria.core.utils import create_thumbnail
+
+        try:
+            create_thumbnail(latest_original)
+        except DjangoCoreValidationError as e:
+            log.error(
+                "Object {obj} created successfully. Thumbnail creation failed. Error: {error}".format(
+                    obj=latest_original, error=e.messages
+                )
+            )
```
fugal-dy, line 181: I think this should respect the `ALEXANDRIA_ENABLE_THUMBNAIL_GENERATION` setting.

alexandria/core/tests/test_models.py
```diff
@@ -1,17 +1,32 @@
...
-from alexandria.core.models import File
+from alexandria.core.models import Document, File
+
+
+def get_latest_file(document):
+    return (
+        document.files.filter(variant=File.Variant.ORIGINAL)
+        .order_by("-created_at")
+        .first()
+    )
```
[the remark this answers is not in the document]
anehx: Good idea. I put it into a method, not a property but if that is ever needed it's a quick change.

alexandria/core/models.py
anehx: Nice catch! I put the condition directly into the `create_thumbnail` function so we don't have to think about it on every callsite.

alexandria/core/utils.py
```diff
@@ -11,6 +11,9 @@
 
 
 def create_thumbnail(file_: File):
+    if not settings.ALEXANDRIA_ENABLE_THUMBNAIL_GENERATION:
+        return
+
```
fugal-dy, line 16: Now that you bring up the function `create_thumbnail`:
 - it kinda looks like this function it's pretty much a method of `File`
 - we could get rid of `utils.py` moving the function to `File` because nothing left to see

anehx: Another fair point, done.

fugal-dy (approved):
good job.

PR #453, opened 2024-02-12
feat: integrate manabi for editing files over WebDAV

open-dynaMIX:


open-dynaMIX:
This PR is very raw and not yet functioning. Inspiration on how to integrate a WSGIDav-app into Django came from [here](https://github.com/qdqmedia/django-templation).

Next steps for this PR:

 - The normal testclient doesn't use the `wsgi.py`. We should look into using something like [webtest](https://github.com/Pylons/webtest) for the webDAV tests
 - MANABI_TOKEN is regenerated on every start. I guess this doesn't work, since the locks are in the DB
 - Make everything run with local filesystem and s3
 - Implement hooks to update `File` in Alexandria, if they are written via webDAV

open-dynaMIX:
In order to easily support encryption and virus scanning with the DAV API, the approach had to be changed:

We don't use `ManabiProvider`/`ManabiS3Provider`  anymore, but a custom `AlexandriaProvider` with its accompanying `AlexandriaFileResource`. This enables us to directly work with the `File` model of Alexandria (that takes care of en- and decryption). This way we also have access to the data, before they are written and can scan for viruses, if this setting is enabled.

alexandria/dav.py
```diff
@@ -0,0 +1,163 @@
...
+from alexandria.core.models import File
+
+
+class MyBytesIO(io.BytesIO):
+    """
+    Custom BytesIO.
+
+    `wsgidav.RequestServer.do_PUT` would close the file after writing. This class just
+    overrides the `close()`-method in order to prevent this, when it's called without
+    `ignore=False`.
```
winged, line 29: Needs more explanation - why is this needed? What would happen without it?

alexandria/dav.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+
+    def close(self, ignore=True):
+        if not ignore:
+            return super().close()
```
winged, line 34:
```suggestion
    def close(self):
        # Prevent close from ...?
        return
    def do_close(self):
        return super().close()
```

alexandria/dav.py
```diff
@@ -0,0 +1,163 @@
...
+                modified_by_user=self.user,
+                modified_by_group=self.group,
+            )
+        file.size = self.memory_file.getbuffer().nbytes
+        self.memory_file.seek(0)
+        django_file = DjangoFile(name=file.name, file=self.memory_file)
+        file.content = django_file
+        file.save()
+        self.file = file
+        self.memory_file.close(ignore=False)
```
winged, line 117:
```suggestion
        self.memory_file.do_close()
```

alexandria/wsgi.py
```diff
@@ -9,8 +9,27 @@
...
-application = get_wsgi_application()
+if settings.ALEXANDRIA_USE_MANABI:
+    from alexandria.dav import get_dav
+
+    wsgi_dav = get_dav()
+    wsgi_django = get_wsgi_application()
+
+    def dispatch(environ, start_response):
+        path = environ["PATH_INFO"] or "/"
+        dav_prefix = "/dav"
```
winged, line 25: Should this path be configurable?

winged (changes requested):
Awesome work! Just some smaller suggestions (debatable), otherwise LGTM

alexandria/dav.py
open-dynaMIX: Done.

alexandria/dav.py
open-dynaMIX: Done.

alexandria/dav.py
open-dynaMIX: Done.

alexandria/wsgi.py
open-dynaMIX: Done.

winged (approved):
Awesome! :tada: :100:

PR #455, opened 2024-02-15
feat(file): add method to generate a download url for a file

anehx:


fugal-dy (approved):
When I first created that serializer field I remeber, that I was considering to make it a model-method as well but chose to include it directly in the serializer because it will always need to be called in context of a request. But I also see the point why you would change it that way.

anehx:
> When I first created that serializer field I remeber, that I was considering to make it a model-method as well but chose to include it directly in the serializer because it will always need to be called in context of a request. But I also see the point why you would change it that way.

I mainly changed it because I had to copy paste the exact same code into a project using alexandria to generate a download link outside of the serializer so I thought this would make sense..

fugal-dy:
Yes, that's what I thought

PR #463, opened 2024-02-20
feat: add optional dms integration

Yelinz:
This adds the feature to be able to send docx, odt to DMS to be converted to pdf and saved in alexandria.


alexandria/core/views.py
```diff
@@ -100,6 +108,44 @@ def update(self, request, *args, **kwargs):
...
+        if response.status_code == HTTP_401_UNAUTHORIZED:
+            raise AuthenticationFailed(_("Token has expired."))
+
+        response.raise_for_status()
+
+        converted_document = models.Document.objects.create(
+            title={k: v + ".pdf" for k, v in document.title.items()},
+            description=document.description,
+            category=document.category,
+            date=document.date,
```
anehx, line 135: `created_by` and `modified_by` are missing

alexandria/core/views.py
```diff
@@ -100,6 +108,44 @@ def update(self, request, *args, **kwargs):
...
+        )
+        file_name = file.name + ".pdf"
+        converted_file = models.File.objects.create(
+            document=converted_document,
+            name=file_name,
+            content=ContentFile(response.content, file_name),
+            mime_type="application/pdf",
+            size=len(response.content),
```
anehx, line 143: Same here

alexandria/core/views.py
```diff
@@ -100,6 +108,44 @@ def update(self, request, *args, **kwargs):
...
+
+        response.raise_for_status()
+
+        converted_document = models.Document.objects.create(
+            title={k: v + ".pdf" for k, v in document.title.items()},
+            description=document.description,
+            category=document.category,
+            date=document.date,
+        )
+        file_name = file.name + ".pdf"
```
anehx, line 137: Does this mean that `foo.docx` will become `foo.docx.pdf`?

Also, same question goes for the document title.

alexandria/core/views.py
```diff
@@ -100,6 +108,44 @@ def update(self, request, *args, **kwargs):
...
+        converted_file = models.File.objects.create(
+            document=converted_document,
+            name=file_name,
+            content=ContentFile(response.content, file_name),
+            mime_type="application/pdf",
+            size=len(response.content),
+        )
+        converted_file.create_thumbnail()
+
+        return Response(status=HTTP_200_OK)
```
anehx, line 147: Would it make sense to return the newly created document? Also, status should be `HTTP_201_CREATED`

alexandria/core/views.py
```diff
@@ -100,6 +108,44 @@ def update(self, request, *args, **kwargs):
...
+
+        response = requests.post(
+            settings.ALEXANDRIA_DMS_URL + "/convert",
+            data={"target_format": "pdf"},
+            headers={"authorization": get_authorization_header(request)},
+            files={"file": file.content},
+        )
+
+        if response.status_code == HTTP_401_UNAUTHORIZED:
+            raise AuthenticationFailed(_("Token has expired."))
```
anehx, line 127: Let's return the error message that the DMS responds with.

anehx (changes requested):

alexandria/core/views.py
Yelinz: Yes, not sure if trying to clean up the file extensions makes sense. As a file name can be anything

winged:
Sorry for being the party crasher here: I feel like maybe this integration going too far. If we add support for all auxiliary services to Alexandria, we'll get quite an untangled mess of (optional) dependencies, not just here but in our other apps too. I'm not generally against this PR, but I think it's useful to reflect on how to do service integration.

Could we possibly integrate things the other way round? Say instead of doing the endpoint in alexandria, how about we support fetching files from Alexandria (or even any endpoint) in the DMS instead? That would allow the DMS to easily integrate even with Caluma for example, without any additional code.

anehx:
> Sorry for being the party crasher here: I feel like maybe this integration going too far. If we add support for all auxiliary services to Alexandria, we'll get quite an untangled mess of (optional) dependencies, not just here but in our other apps too. I'm not generally against this PR, but I think it's useful to reflect on how to do service integration.

We did discuss this in length in the concept phase of this feature. I understand your concern but as it is completely opt-in I don't see the issue. The alternative would be to implement a PDF conversion in alexandria which would add a huge dependency named unoconv / libreoffice. I don't think this is the better option.

> Could we possibly integrate things the other way round? Say instead of doing the endpoint in alexandria, how about we support fetching files from Alexandria (or even any endpoint) in the DMS instead? That would allow the DMS to easily integrate even with Caluma for example, without any additional code.

No we could not, as this is a feature which will be used in the frontend of alexandria. Why would the frontend of alexandria use DMS?

For context: the requirement is that alexandria files of the type docx or odt can be easily converted to PDF. The file should automatically be created in the same category.

alexandria/core/views.py
```diff
@@ -100,6 +108,44 @@ def update(self, request, *args, **kwargs):
 
         return response
 
+    @action(methods=["get"], detail=True)
```
anehx, line 111: Needs to be post

winged:
Had a discussion with @anehx just before

> We did discuss this in length in the concept phase of this feature. I understand your concern but as it is completely opt-in I don't see the issue. The alternative would be to implement a PDF conversion in alexandria which would add a huge dependency named unoconv / libreoffice. I don't think this is the better option.

> For context: the requirement is that alexandria files of the type docx or odt can be easily converted to PDF. The file should automatically be created in the same category.

I see - the actual use case wan't entirely clear in the PR or commit message. The way the feature is intended to work, I also can't propose a cleaner way to do it. Some ways with more separation would just move the complexity around and possibly make it much worse.

So, I'll retract my rejection of the PR here - feel free to merge (after @anehx feedback is applied ofc). The general discussion about service integration will take place unrelated to this.

alexandria/core/views.py
Yelinz: using [splitext](https://docs.python.org/3/library/os.path.html#os.path.splitext) for cleaning

alexandria/core/views.py

alexandria/core/views.py

alexandria/core/views.py
Yelinz: Returning 201 with the new pdf

alexandria/core/views.py
Yelinz: Done

alexandria/core/views.py
Yelinz: Done

alexandria/core/views.py
anehx: DMS will return JSON, no?

alexandria/core/views.py
Yelinz: Ok returning serialized json

alexandria/core/views.py
Yelinz: Parsing the response now

anehx (approved):

anehx:
LGTM! @winged could you do a quick review as well?

PR #470, opened 2024-02-29
fix(webdav): fix anonymous user attributes

Yelinz:
use the provided helper function to extract the attributes, which checks for anonymous aswell


Yelinz:
yes forgot formatting

PR #471, opened 2024-02-29
Various fixes for the WebDAV-API

open-dynaMIX:
- fix: only serve newest File form Document
  This is needed for multiple reasons:

   1. There is no need to edit old files and the result would probably be
      unexpected
   2. When creating a new version, Libreoffice doesn't know about this and
      it would use the older File as startingpoint. Leading to a new
      version on every save.
 - fix: only create webDAV links for original Files
 - fix: run hooks after file save was successful
 - chore: simplify fetching user attributes from request

### DIFF
--- a/alexandria/core/serializers.py
+++ b/alexandria/core/serializers.py
@@ -182,19 +182,15 @@ class FileSerializer(BaseSerializer):
         request = self.context.get("request")
         scheme = request.scheme if request else "http"
         host = request.get_host() if request else "localhost"
-        username = (
-            default_user_attribute(
-                request.user, settings.ALEXANDRIA_CREATED_BY_USER_PROPERTY
-            )
-            if request is not None
-            else None
+        username = getattr(
+            getattr(request, "user", None),
+            settings.ALEXANDRIA_CREATED_BY_USER_PROPERTY,
+            None,
         )
-        group = (
-            default_user_attribute(
-                request.user, settings.ALEXANDRIA_CREATED_BY_GROUP_PROPERTY
-            )
-            if request is not None
-            else None
+        group = getattr(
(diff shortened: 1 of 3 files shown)


alexandria/core/serializers.py
```diff
@@ -15,7 +15,11 @@
 
 
 def default_user_attribute(user, attribute):
-    return getattr(user, attribute) if not isinstance(user, AnonymousUser) else None
+    return (
+        getattr(user, attribute)
+        if user is not None and not isinstance(user, AnonymousUser)
+        else None
+    )
```
winged, line 22: Could we do this instead?

```suggestion
    return getattr(user, attribute, None)
```

winged (approved):
LGTM, just a suggestion to simplify the `default_user_attribute()` thingy a bit (not 100% sure it fully covers our needs, but should be logically equivalent)

alexandria/core/serializers.py
winged: This will also just return `None` if you mistype an attribute on a regular user object, so maybe not the best idea, but we'd fail hard in that case anyway

open-dynaMIX: I went a different path. I don't feel like there needs to be a function that does nothing more than `getattr`.

winged (commented):
Nice! But I think it would make sense to get rid of the `default_user_attribute()` function entirely (ideally in a separate PR or at least in it's own commit)

open-dynaMIX:
Agree. I've added a commit for that.

> Nice! But I think it would make sense to get rid of the `default_user_attribute()` function entirely (ideally in a separate PR or at least in it's own commit)

I agree. For keeping it simple, I've added a commit for that to this PR.

PR #487, opened 2024-03-13
feat: add utils for creating files

Yelinz:
avoids forgetting to set encryption_status


czosel (approved):

alexandria/core/utils.py
```diff
@@ -0,0 +1,41 @@
+from django.conf import settings
+from django.core.exceptions import ValidationError
+
+from alexandria.core import models
+
+
+def create_document_file(user, group, document_attributes, file_attributes):
```
winged, line 7: Please add a docstring, as these functions will likely be "public API"

alexandria/core/utils.py
```diff
@@ -0,0 +1,41 @@
...
+        modified_by_user=user,
+        modified_by_group=group,
+        **document_attributes,
+    )
+    file = create_file(document, user, group, file_attributes)
+
+    return document, file
+
+
+def create_file(document, user, group, attributes):
```
winged, line 20: Same here, please add a docstring

winged (changes requested):
Code LGTM, but a few things to note:

* Rename the `utils.py` to `api.py` (similar to Caluma, where the public python API is also in an `api` module)
* Add some docstrings to explain what the functions do
* Add some documentation (README.md or create a new file PYTHON_API.md or similar)

alexandria/core/utils.py
Yelinz: Add reST docs or was the idea something different

alexandria/core/utils.py
Yelinz: Add reST docs or was the idea something different

PYTHON_API.md
```diff
@@ -0,0 +1,25 @@
...
+`create_file`
+Creates a file with thumbnail
+Params:
+
+- `document` The document associated with the file.
+- `user` The user who created the file.
+- `group` The group who created the file.
+- `attributes` A dictionary containing the fields for the file. (required: name, content, mime_type, size)
+  Return:
+- The created file object.
```
czosel, line 25: can you add a link to this file in README.md?

alexandria/core/api.py
```diff
@@ -0,0 +1,66 @@
...
+        created_by_group=group,
+        modified_by_user=user,
+        modified_by_group=group,
+        **attributes,
+    )
+
+    try:
+        file.create_thumbnail()
+    except ValidationError:  # pragma: no cover
+        # thumbnail could not be generated because of an unsupported mime type
```
winged, line 63: We're ignoring exceptions here, but in the REST API, we return an error. I think we should be consistent in behaviour between REST and Python API

alexandria/core/api.py
```diff
@@ -0,0 +1,66 @@
...
+        modified_by_user=user,
+        modified_by_group=group,
+        **document_attributes,
+    )
+    file = create_file(document, user, group, file_attributes)
+
+    return document, file
+
+
+def create_file(document, user, group, attributes):
```
winged, line 33: I'd like to see `**kwargs` instead of passing around dicts, at least for the optional params. The required ones should be regular args without defaults

alexandria/core/api.py
```diff
@@ -0,0 +1,66 @@
+from django.conf import settings
+from django.core.exceptions import ValidationError
+
+from alexandria.core import models
+
+
+def create_document_file(user, group, document_attributes, file_attributes):
```
winged, line 7: Since there are some required arguments, it may make sense to have them listed directly in the arguments. That way, it's clear without even needing to document them.

The file arguments could be renamed, so as to show which ones are required as well. Maybe something like this:

```suggestion
def create_document_file(
    user,
    group,
    category,
    title,
    file_name,
    file_content,
    mime_type,
    file_size,
    additional_attributes=None,
    additional_file_attributes=None
):
    additional_attributes = additional_attributes or {}
    additional_file_attributes = additional_file_attributes or {}
```

alexandria/core/api.py
```diff
@@ -0,0 +1,66 @@
...
+    Create a document and file with the given attributes.
+
+    This function eases the creation of documents and files by automatically setting important fields.
+    Uses ``create_file`` to create the file.
+
+    :param user: The user creating the document and file.
+    :param group: The group creating the document and file.
+    :param document_attributes: A dictionary containing the fields for the document. (required: category, title)
+    :param file_attributes: A dictionary containing the fields for the file. (required: name, content, mime_type, size)
+    :return: A tuple containing the created document and file.
```
winged, line 18: I think the reST syntax is obsolete. Instead, use type annotations where possible, and just plain text description of the arguments. For example here, it's not clear whether the user and group params are strings or objects.

PYTHON_API.md
```diff
@@ -0,0 +1,25 @@
...
+- A tuple containing the created document and file.
+
+`create_file`
+Creates a file with thumbnail
+Params:
+
+- `document` The document associated with the file.
+- `user` The user who created the file.
+- `group` The group who created the file.
+- `attributes` A dictionary containing the fields for the file. (required: name, content, mime_type, size)
```
winged, line 23: What types are those parameters? Especially looking at `content`

winged (changes requested):
Thought about the whole thing some more, and I think for a "public" API, we need to refine it a bit more. Good work so far though!

winged:
> Thought about the whole thing some more, and I think for a "public" API, we need to refine it a bit more. Good work so far though!

Yet another thought: some of the functionality could probably be put into django signal handlers instead (`before_save` for determining a file's size, `post_save` for creating the thumbnail).

Yelinz:
> > Thought about the whole thing some more, and I think for a "public" API, we need to refine it a bit more. Good work so far though!
> 
> Yet another thought: some of the functionality could probably be put into django signal handlers instead (`before_save` for determining a file's size, `post_save` for creating the thumbnail).

Could be considered for a seperate PR, as that would introduce some more code overhaul. We can open an issue for this.
Changed the api to be more explicit, previously I kept it with dicts, to be more flexible(`category` or `category_id`).

alexandria/core/api.py
Yelinz: On the REST API we also except the error, but generate an `log.error` as well. Should the python api also `log.error` in that case?

alexandria/core/views.py
```diff
@@ -136,31 +137,25 @@ def convert(self, request, pk=None):
...
+            "application/pdf",
+            len(response.content),
+            {
+                "description": document.description,
+                "date": document.date,
+                "metainfo": document.metainfo,
+            },
+            {
+                "metainfo": file.metainfo,
+            },
```
winged, line 157: please use named arguments when there are more than 2-3 parameters. It becomes too easy to make mistakes otherwise

alexandria/core/api.py
Yelinz: Added error logging

PYTHON_API.md
Yelinz: Done

PYTHON_API.md
```diff
@@ -0,0 +1,35 @@
...
+- `document` The document associated with the file.
+- `user` The user who created the file.
+- `group` The group who created the file.
+- `name` Name of the file
+- `content` File object
+- `mime_type` File mime type
+- `size` File size
+- `attributes` Kwargs containing the optional fields for the file. 
+  Return:
+- The created file object.
```
winged, line 35: This should be reformatted so it looks good rendered as HTML (by Github) as well. That's not quite the case yet.

alexandria/core/api.py
winged: I would suggest renaming the `**attributes` / `**file_attributes` to `**additional_attributes`, to further show that they're *additional* to the regular ones.

Otherwise, the PR looks good now, and ready to merge

winged (approved):

winged:
Nice work! Please squash before merge though, we don't need all intra-PR fixes in the history

PR #490, opened 2024-03-15
fix(dav): set created_by_* attributes on new files

open-dynaMIX:


alexandria/dav_provider.py
```diff
@@ -112,6 +112,8 @@ def end_write(self, *, with_errors):
                 document=self.file.document,
                 encryption_status=self.file.encryption_status,
                 mime_type=self.file.mime_type,
+                created_by_user=self.user,
+                created_by_group=self.group,
```
ganwell, line 116: That looks very fishy too me. In Camac-Bern, the version log worked correctly when not hijacking files created by other users. And just setting modiffied_by_user.

open-dynaMIX: We need to set those properties on new files. This was forgotten. Not sure if it fixes the issue at hand, but it needs to be done anyway. In the rest-API this is done in the serializer: https://github.com/projectcaluma/alexandria/blob/main/alexandria/core/serializers.py#L36

czosel (approved):

PR #497, opened 2024-03-19
fix(upload): use content-type and filename to infer mime types

czosel:
This fixes an issue we saw when uploading docx documents generated by document-merge-service. We were previously relying solely on content-based mime type detection (using python-magic), which caused an incorrect detection for such documents as ZIP files instead of docx. This change makes use of the type information we have available:

1. Content-type header
2. If it's application/octet-stream, guess mime type from file extension
3. If no guess is available, content-based guessing based (as before)


alexandria/core/serializers.py
```diff
@@ -219,6 +221,12 @@ def validate(self, *args, **kwargs):
                 f'"original" must not be set for variant "{variant}".'
             )
 
+        mime_type = validated_data["content"].content_type
+        if mime_type == "application/octet-stream":
+            guess, _ = guess_type(validated_data["name"])
+            if guess is not None:
+                mime_type = guess
+        validated_data["mime_type"] = mime_type
         validated_data["mime_type"] = validated_data["content"].content_type
```
Yelinz, line 230: Is this intended to be left? doesnt this overwrite the changes just made

czosel: Oops that is a leftover indeed

PR #500, opened 2024-03-20
Cleanup thumbnail creation

czosel:


alexandria/core/serializers.py
```diff
@@ -199,35 +198,27 @@ def get_webdav_url(self, instance):
         )
         return instance.get_webdav_url(username, group, host)
 
+    def validate_variant(self, value):
+        if not value:
+            value = "original"
```
anehx, line 203: Isn't it required anyways?

alexandria/core/serializers.py
```diff
@@ -199,35 +198,27 @@ def get_webdav_url(self, instance):
...
+        if value != "original":
```
anehx, line 204:
```suggestion
        if value != models.File.Variant.ORIGINAL:
```

alexandria/core/serializers.py
```diff
@@ -199,35 +198,27 @@ def get_webdav_url(self, instance):
...
+            raise serializers.ValidationError("only original files can be created")
```
anehx, line 205: Translate

alexandria/core/serializers.py
```diff
@@ -273,7 +264,6 @@ class Meta:
         fields = BaseSerializer.Meta.fields + (
             "variant",
             "name",
-            "original",
```
anehx, line 276: Revert. Otherwise it will not appear in the response as seen in the snapshots.

Read only fields must be declared in both `fields` **and** `read_only_fields`

anehx (changes requested):

alexandria/core/serializers.py
czosel: simplified in other PR: `return "original"`

alexandria/core/serializers.py
czosel: removed in other PR

alexandria/core/serializers.py
czosel: good catch, fixed in other PR.

czosel:
Closed in favor of #501

fugal-dy (approved):
okay. nice and clean

PR #501, opened 2024-03-20
chore: remove thumbnail creation via API

czosel:
This removes support for "manually" uploading thumbnails via the API, because there is no use-case for this (thumbnails should always be created server-side).
Also move the file `create` function from the view to the serializer.


alexandria/core/serializers.py
```diff
@@ -199,27 +199,20 @@ def get_webdav_url(self, instance):
...
-
-        if (
-            variant := validated_data.get("variant")
-        ) == models.File.Variant.ORIGINAL and validated_data.get("original"):
-            raise ValidationError(
-                f'"original" must not be set for variant "{variant}".'
-            )
 
         mime_type = validated_data["content"].content_type
         if mime_type == "application/octet-stream":
```
fugal-dy, line 218: the line below should be removed again `validated_data["mime_type"] = validated_data["content"].content_type`

anehx: https://github.com/projectcaluma/alexandria/pull/499

czosel: This is done in another PR: #499

anehx (approved):
Nice cleanup! :100: 

Did you test it with the frontend?

alexandria/core/views.py
```diff
@@ -244,43 +244,6 @@ def multi(self, request, **kwargs):
 
             return response
 
-    def create(self, request, *args, **kwargs):
-        # invoke dgap permission check
-        self._check_permissions(request)
```
Yelinz, line 249: Should not be removed afaik

Yelinz (changes requested):

czosel:
@Yelinz I noticed that as well, fix is here: https://github.com/adfinis/django-generic-api-permissions/releases/tag/v0.4.4

czosel:
bumped dgap, manual testing with frontend works now

Yelinz (approved):

PR #507, opened 2024-03-26
feat(document)!: create document and file in one request!

Yelinz:
BREAKING CHANGE:
The document post endpoint now requires the file data to be provided as well.
The reason for this change is allowing the frontend to create documents
and files in one request, preventing documents with no associated files.
Which fixes the problem if the file got rejected for any reason, the
application would create an empty document.

### DIFF
--- a/alexandria/core/tests/test_permissions.py
+++ b/alexandria/core/tests/test_permissions.py
@@ -33,6 +33,7 @@ def test_document_permission(
     admin_user,
     admin_client,
     client,
+    document_post_data,
     method,
     status,
     use_admin_client,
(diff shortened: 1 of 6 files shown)


alexandria/core/tests/test_permissions.py
```diff
@@ -61,25 +63,27 @@ def has_object_permission_for_document(self, request, instance):
...
+                    "category": {"data": {"id": doc.category.pk, "type": "categories"}}
+                },
+            }
+        },
+        "delete": None,
+    }
 
-    response = getattr(client, method)(url, data=data)
+    response = getattr(client, method)(
+        url, data=data[method], format=None if method != "post" else "multipart"
```
winged, line 85: Ternary could be inverted for readability

```suggestion
        url, data=data[method], format="multipart" if method == "post" else None
```

winged (changes requested):
Code LGTM, apart from a tiny readability issue.

And I'd like to see a bit of an explanation why this is needed in the commit message / PR description.

alexandria/core/tests/test_permissions.py
Yelinz: applied

winged (approved):

PR #513, opened 2024-04-03
fix: move file validation to model

czosel:
This way, the same validations are also triggered when files are uploaded via the python API.

### DIFF
--- a/alexandria/core/models.py
+++ b/alexandria/core/models.py
@@ -181,8 +181,6 @@ class Document(UUIDModel):
                 latest_original.document = self
                 latest_original.save()
 
-        latest_original.create_thumbnail()
-
         return self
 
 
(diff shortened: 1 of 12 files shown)


alexandria/core/models.py
```diff
@@ -239,8 +241,24 @@ def set_checksum(self):
     def save(
         self, force_insert=False, force_update=False, using=None, update_fields=None
     ):
+        validate_file_infection(self.content)
```
winged, line 244: I'm not too sure about having this level of validation it in the model. Could we move it to a DGAP validator and trigger that from the Python API instead? Of course that would imply applications need to be aware of the DGAP class doing this, but still, I think it'd be worth it for a cleaner codebase

Yelinz: Created a base validator class which is passed to dgap per default.

winged (approved):
LGTM now I think

PR #529, opened 2024-04-19
File creation fixes

Yelinz:
- fixes non atomic creation of documents
- allows file extension checking


alexandria/core/validations.py
```diff
@@ -40,6 +40,9 @@ def validate_file(self, data, context):
             guess, encoding = guess_type(data["name"])
             if guess is not None:
                 mime_type = guess
+            else:
+                # use file extension if no mime type is found (e.g. for .msg files)
+                mime_type = data["name"].split(".")[-1]
```
winged, line 45: I'd rather fall back to a custom mapping, and when that fails, "application/octet-stream". Or do we have a concrete case where browsers fail to provide the proper mime type at all, and it's business-critical to accept a given file?

winged: Alternatively, there's also [libmagic](https://pypi.org/project/python-magic/) that can look at the content of a file instead of the extension

winged (changes requested):

alexandria/core/validations.py
Yelinz: Makes sense to use libmagic, as it is already installed for the preview generator

Yelinz: Or not, lets just do a custom mapping

Yelinz: Hmm, we have no file descriptor at that point, as we are working with the bytes content from the response.

winged: I'd still say go for it. Uploads don't happen by "anonymous" users, so I think the risk of DoS is rather limited. And we can deal with a few memory spikes if some people upload bigger documents (IMHO). And it would only trigger if the by-extension guessing code fails, so fairly seldom

Yelinz: Well tried a little bit, but wasnt able to get it to work with file descriptors without creating a temporary file. I dont think that is a good idea.

winged (approved):

PR #535, opened 2024-04-30
feat: allow custom implementation of username and group getters

czosel:
This adds more flexibility when alexandria is used as a package. The consuming app can dictate in more fine grained detail how user and group are set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ A list of configuration options which you need
   - `ALEXANDRIA_OIDC_USER_FACTORY`: Overwrite the default user with your own
   - `ALEXANDRIA_CREATED_BY_USER_PROPERTY`: Overwrite the default user property which is used for `..._by_user` (default: username)
   - `ALEXANDRIA_CREATED_BY_GROUP_PROPERTY`: Overwrite the default group property which is used for `..._by_group` (default: group)
+  - `ALEXANDRIA_GET_USER_AND_GROUP_FUNCTION`: Overwrite the import string if further configuration how user and group are determined is needed.
 - Authorization configurations
   - `ALEXANDRIA_VISIBILITY_CLASSES`: Comma-separated list of [DGAP](https://github.com/adfinis/django-generic-api-permissions/?tab=readme-ov-file#visibilities) classes that define visibility for all models
   - `ALEXANDRIA_PERMISSION_CLASSES`: Comma-separated list of [DGAP](https://github.com/adfinis/django-generic-api-permissions/?tab=readme-ov-file#permissions) classes that define permissions for all models
--- a/alexandria/core/serializers.py
+++ b/alexandria/core/serializers.py
@@ -5,6 +5,7 @@ from django.conf import settings
 from django.db.transaction import atomic
 from django.template.defaultfilters import slugify
 from django.utils import translation
+from django.utils.module_loading import import_string
 from generic_permissions.validation import ValidatorMixin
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -73,6 +73,7 @@ A list of configuration options which you need
   - `ALEXANDRIA_OIDC_USER_FACTORY`: Overwrite the default user with your own
   - `ALEXANDRIA_CREATED_BY_USER_PROPERTY`: Overwrite the default user property which is used for `..._by_user` (default: username)
   - `ALEXANDRIA_CREATED_BY_GROUP_PROPERTY`: Overwrite the default group property which is used for `..._by_group` (default: group)
+  - `GET_USER_AND_GROUP(request)`: Overwrite the extraction function if further configuration how user and group are determined.
```
czosel, line 76:
```suggestion
  - `GET_USER_AND_GROUP(request)`: Overwrite the extraction function if further configuration how user and group are determined is needed. 
```

czosel:
Otherwise LGTM as well :+1:

winged:
Just thought about it some more. I think it would be better to have the *setting* point, as a string, to the actual function. The *caller* (in the serializer) would then use `django.utils.module_loading.import_string()` to get the actual function.

This way, it remains a setting that could even be overloaded via env vars, and not some special case where you'd need to inject code into the settings module

czosel:
> Just thought about it some more. I think it would be better to have the _setting_ point, as a string, to the actual function. The _caller_ (in the serializer) would then use `django.utils.module_loading.import_string()` to get the actual function.
> 
> This way, it remains a setting that could even be overloaded via env vars, and not some special case where you'd need to inject code into the settings module

Fair point, that would be closer to what we do in Caluma as well..

Yelinz:
> Just thought about it some more. I think it would be better to have the _setting_ point, as a string, to the actual function. The _caller_ (in the serializer) would then use `django.utils.module_loading.import_string()` to get the actual function.
> 
> This way, it remains a setting that could even be overloaded via env vars, and not some special case where you'd need to inject code into the settings module

Refactored to be an import string

alexandria/core/serializers.py
```diff
@@ -15,6 +16,7 @@
 from . import models
 
 log = logging.getLogger(__name__)
+get_user_and_group = import_string(settings.ALEXANDRIA_GET_USER_AND_GROUP_FUNCTION)
```
winged, line 19: I'd move it into a method of the `BaseSerializer`. If you have it here, it's hard to change during tests, as it won't be re-evaluated if you change the setting. Or keep it just inline where it's being called instead.

winged (changes requested):
I like it much better! Just the scoping should be improved in `serializers.py`

alexandria/core/serializers.py
```diff
@@ -183,7 +186,9 @@ def get_webdav_url(self, instance):
             return None
         request = self.context.get("request")
         host = request.get_host() if request else "localhost"
-        user, group = get_user_and_group(request)
+        user, group = import_string(settings.ALEXANDRIA_GET_USER_AND_GROUP_FUNCTION)(
+            self.context["request"]
+        )
```
winged, line 191: I think this could be factored out into a method here - same (relatively-complex) code pattern 3 times in the same class..

Yelinz: Makes sense, done

winged (approved):
Nice work! LGTM

PR #537, opened 2024-05-01
File content search

Yelinz:
Implement file content search with tika and psql full text search


alexandria/settings/alexandria.py
```diff
@@ -194,3 +194,15 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
     "MANABI_SHARED_KEY", default=default("2QhWg20fXq0xlnJUkFTDgFoA3JWqvb86OejD9mAVFCW")
 )
 MANABI_SECURE = True if ENV == "production" else False
+
+
+# Content search
+ALEXANDRIA_ENABLE_CONTENT_SEARCH = env.bool(
```
Yelinz, line 200: Should this feature be configurable to be on or off?

winged: I think it should be default-off, as it requires additional infrastructure and CPU load

alexandria/core/filters.py
```diff
@@ -132,6 +136,25 @@ def filter(self, qs, value):
...
+        search_configs = {
+            settings.ISO_639_TO_PSQL_SEARCH_CONFIG.get(lang, "simple")
+            for lang, desc in settings.LANGUAGES
+        }
+
+        search_queries = [
+            SearchQuery(value, config=config, search_type="phrase")
+            for config in search_configs
+        ]
+        search_query = reduce(or_, search_queries)
```
winged, line 153: I think you can improve the performance quite a bit if you pair each `SearchQuery` with an additional `Q` object that matches the file's detected language. That way, you won't search german-stemmed expressions in english text, which also has the benefit of avoiding possible false positives

alexandria/core/filters.py
```diff
@@ -164,6 +187,19 @@ class FileFilterSet(FilterSet):
...
+    latest = BooleanFilter(method="filter_latest")
+
+    def filter_latest(self, qs, name, value):
+        if not value:
+            return qs
+
+        latest_files = models.File.objects.filter(
+            document=OuterRef("document_id")
+        ).order_by("-created_at")
+        latest_file_of_documents = qs.filter(pk=Subquery(latest_files.values("pk")[:1]))
```
winged, line 200: Can you try and see what the actual SQL query comes out of this? I'm afraid this might not perform very well.

If neccessary, I think we should add an `is_latest` field to `File` that's updated when new versions are added..

But then again, maybe we need to iterate on this anyway: We want the latest version with a *match* in the search, which may not be in the latest actual version of the file...

alexandria/core/models.py
```diff
@@ -220,6 +223,9 @@ class Variant(models.TextChoices):
         blank=True,
     )
 
+    content_vector = SearchVectorField(null=True, blank=True)
```
winged, line 226: You need to add an explicit index on this field, otherwise the search will not perform much better than any naive `__icontains` lookup

winged (changes requested):
Awesome work so far! I think it's going into the right direction.

A few general hints:

* Please split up the work into a few more commits (infra vs model vs filters), and add some more description as to what/how/why things are happening.
* Extend the README to explain the feature, not just where the flag is enabled
* I think we wanted to create a separate endpoint for the search - That way, we can also expose [search headline](https://docs.djangoproject.com/en/5.0/ref/contrib/postgres/search/#searchheadline) and easily order by search rank. You can then add a related-field to it and even include the actual file object

And some more ideas / brainstorming:

* Add filename / document title as weighted search vector onto the field as well. That way, users can search by file name and will find a match ([see here](https://docs.djangoproject.com/en/5.0/ref/contrib/postgres/search/#weighting-queries))
*

alexandria/core/filters.py
Yelinz: In a previous iteration of the filter I was generating the subquery with
```py
qs.filter(document=OuterRef("document_id")).order_by("-created_at")
```
This was filtering for the latest matching file.

The current query filters for the latest file of the document.
Django generated output:
```sql
SELECT
    "alexandria_core_file"."created_at",
    "alexandria_core_file"."created_by_user",
    "alexandria_core_file"."created_by_group",
    "alexandria_core_file"."modified_at",
    "alexandria_core_file"."modified_by_user",
    "alexandria_core_file"."modified_by_group",
    "alexandria_core_file"."metainfo",
    "alexandria_core_file"."id",
    "alexandria_core_file"."variant",
    "alexandria_core_file"."original_id",
    "alexandria_core_file"."name",
    "alexandria_core_file"."document_id",
    "alexandria_core_file"."checksum",
    "alexandria_core_file"."encryption_status",
    "alexandria_core_file"."content_vector",
    "alexandria_core_file"."language",
    "alexandria_core_file"."content",
    "alexandria_core_file"."mime_type",
    "alexandria_core_file"."size"
FROM
    "alexandria_core_file"
WHERE
    (
        "alexandria_core_file"."variant" = 'original'
        AND "alexandria_core_file"."id" = (
            SELECT
                U0."id"
            FROM
                "alexandria_core_file" U0
            WHERE
                U0."document_id" = "alexandria_core_file"."document_id"
            ORDER BY
                U0."created_at" DESC
            LIMIT
                1
        )
    )
ORDER BY
    "alexandria_core_file"."created_at" DESC;
```

winged: OK, subquery looks good to me, but it's missing the `variant` filter. And thumbnails tend to be newer than the originals, so this will likely break your assumptions

alexandria/core/filters.py
Yelinz: Done

alexandria/core/filters.py
Yelinz: Done

alexandria/core/models.py
Yelinz: Done

alexandria/core/views.py
```diff
@@ -259,3 +263,36 @@ def download(self, request, pk=None):
...
+        latest_files = models.File.objects.filter(
+            document=OuterRef("document_id"), variant=models.File.Variant.ORIGINAL
+        ).order_by("-created_at")
+        queryset = queryset.filter(pk=Subquery(latest_files.values("pk")[:1])).filter(
+            search_query
+        )
+
+        serializer = self.get_serializer(queryset, many=True)
+        headers = self.get_success_headers(serializer.data)
+        return Response(serializer.data, status=HTTP_200_OK, headers=headers)
```
winged, line 298: Code functionally LGTM, but I suggest keeping it in the "traditional" `FilterSet` structure.
This implies adding a new (readonly) viewset, with a new serializer that also exposes the headline / search rank as well.

winged (dismissed):
Nice work so far! I think mostly it's about code organization right now

alexandria/core/views.py
Yelinz: Changed to a filter method. Can be refactored to new viewset when integration additional full text search features.

alexandria/core/management/commands/generate_content_vectors.py
```diff
@@ -0,0 +1,28 @@
...
+                )
+            )
+
+        for file in tqdm(
+            File.objects.filter(variant="original", content_vector__isnull=True)
+        ):
+            file.set_content_vector()
+            file.save()
+
+            if virtual_memory().available < 300000000:  # pragma: no cover
```
winged, line 26:
```suggestion
            if virtual_memory().available < 300_000_000:  # pragma: no cover
```

for readability

alexandria/core/filters.py
```diff
@@ -164,6 +168,33 @@ class FileFilterSet(FilterSet):
     document_metainfo = JSONValueFilter(field_name="document__metainfo")
     active_group = ActiveGroupFilter()
     files = BaseCSVFilter(field_name="pk", lookup_expr="in")
+    content_search = CharFilter(method="filter_content_search")
+
+    def filter_content_search(self, queryset, name, value):
+        if value in EMPTY_VALUES:  # pragma: no cover
+            return queryset
```
winged, line 175: Should not be needed

alexandria/core/management/commands/generate_content_vectors.py
Yelinz: Done

alexandria/core/filters.py
Yelinz: Removed

alexandria/core/views.py
winged: I would suggest extending it to a new viewset (and endpoint) now, as it will be a breaking change if you do it later. And I think the headline/rank will be added relatively quickly

Yelinz: Ok ive created a view for it.

anehx:
What's the status for this feature?

Yelinz:
Works, just needs a review approval. Ill assign it

alexandria/core/factories.py
```diff
@@ -160,7 +160,11 @@ class FileFactory(BaseFactory):
             document=factory.SelfAttribute("..document"),
         ),
     )
-    mime_type = Faker("mime_type")
+    mime_type = factory.Maybe(
+        factory.LazyAttribute(lambda o: o.variant == models.File.Variant.THUMBNAIL),
+        yes_declaration="image/jpeg",
+        no_declaration="image/png",
```
anehx, line 166:
```suggestion
        no_declaration=Faker("mime_type"),
```

alexandria/core/views.py
```diff
@@ -280,3 +284,42 @@ def check_object_permissions(self, request, instance):
...
+            )
+
+        search_query = reduce(or_, search_queries)
+
+        latest_files = models.File.objects.filter(
+            document=OuterRef("document_id"), variant=models.File.Variant.ORIGINAL
+        ).order_by("-created_at")
+        queryset = queryset.filter(pk=Subquery(latest_files.values("pk")[:1])).filter(
+            search_query
+        )
```
anehx, line 323: Why do we need a subquery?

Yelinz: Current the search only searches over latest original files. This subquery is for that.

anehx: Aaah now I get it.. Nevermind.

alexandria/core/factories.py
Yelinz: Done and rebased

anehx:
LGTM, now we should cleanup the git history in order to produce a readable changelog. I'd suggest soft resetting the whole branch and then group the logic into meaningful commits.

anehx (approved):
:tada:

PR #544, opened 2024-05-06
feat(category): add the sort property on the category model

JohnL17:


alexandria/core/models.py
```diff
@@ -117,6 +117,7 @@ class Category(SlugModel):
         null=True,
         related_name="children",
     )
+    sort = models.IntegerField(blank=True, null=True)
```
Yelinz, line 120: Would make sense to add sort as the default ordering to meta

anehx: Done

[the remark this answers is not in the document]
anehx: Done

winged (approved):
LGTM, nice work!

JohnL17:
Whoa I totally forgot that this exists :sweat_smile: thank you all for solving it!

PR #549, opened 2024-05-16
feat(dav)!: dav url to sperate view

Yelinz:
REFACTOR: remove dav url from file serializer

### DIFF
--- a/alexandria/core/serializers.py
+++ b/alexandria/core/serializers.py
@@ -317,9 +317,6 @@ class WebDAVSerializer(BaseSerializer):
     webdav_url = serializers.SerializerMethodField()
 
     def get_webdav_url(self, instance):
-        if not settings.ALEXANDRIA_USE_MANABI:
-            return None
-
         request = self.context.get("request")
         host = request.get_host() if request else "localhost"
         user, group = self.get_user_and_group()
--- a/alexandria/core/serializers.py
+++ b/alexandria/core/serializers.py
@@ -167,7 +167,6 @@ class FileSerializer(BaseSerializer):
     }
 
     download_url = serializers.SerializerMethodField()
-    webdav_url = serializers.SerializerMethodField()
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -178,17 +177,6 @@ class FileSerializer(BaseSerializer):
 
     def get_download_url(self, instance):
         return instance.get_download_url(self.context.get("request"))
-
-    def get_webdav_url(self, instance):
-        if (
-            instance.variant != models.File.Variant.ORIGINAL
-            or not settings.ALEXANDRIA_USE_MANABI
-        ):
-            return None
-        request = self.context.get("request")
-        host = request.get_host() if request else "localhost"
-        user, group = self.get_user_and_group()
-        return instance.get_webdav_url(user, group, host)
 
     def validate(self, *args, **kwargs):
         """Validate the data.
@@ -246,7 +234,6 @@ class FileSerializer(BaseSerializer):
             "checksum",
             "content",
             "download_url",
-            "webdav_url",
             "mime_type",
             "size",
         )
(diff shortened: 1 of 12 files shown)


alexandria/core/serializers.py
```diff
@@ -324,3 +311,20 @@ class Meta:
...
         )
         extra_kwargs = {"content": {"write_only": True}}
+
+
+class WebDAVSerializer(BaseSerializer):
+    webdav_url = serializers.SerializerMethodField()
+
+    def get_webdav_url(self, instance):
+        if not settings.ALEXANDRIA_USE_MANABI:
+            return None
```
open-dynaMIX, line 321: Should we issue a 400 or 404 here? Or is the frontend determining if manabi is used only by receiving `null` here?

open-dynaMIX (approved):
Great work :rocket:

I just have two questions.

alexandria/core/views.py
```diff
@@ -259,3 +193,79 @@ def download(self, request, pk=None):
                 obj.content.file.file, as_attachment=False, filename=obj.name
             )
         raise PermissionDenied("For downloading a file use the presigned download URL.")
+
+
+class DocumentViewSet(PermissionViewMixin, VisibilityViewMixin, ModelViewSet):
```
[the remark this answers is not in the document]
anehx: Had the same question. If not necessary, please move it back..

alexandria/core/serializers.py
anehx: IMO we should throw a 404. Also, could you please add a setting e.g. `ALEXANDRIA_MANABI_ALLOWED_MIMETYPES` with default to docx and xlsx and throw a 404 if the mimetype of the file doesn't match?

I discussed this with @czosel and we both agree that we need such an option because of security reasons of various customers. We don't want to expose all files to webdav.

alexandria/core/serializers.py
```diff
@@ -324,3 +311,20 @@ class Meta:
...
+
+        request = self.context.get("request")
+        host = request.get_host() if request else "localhost"
+        user, group = self.get_user_and_group()
+        return instance.get_latest_original().get_webdav_url(user, group, host)
```
anehx, line 326: Where do we check the permission?

open-dynaMIX: This should be done in the `ViewSet`. It inherits from `PermissionViewMixin`. But I guess this doesn't work like that, because we're making `GET` requests and are not creating or updating any resources in the POV of the rest api.

Yelinz: Permissions are checked with DGAP, with the Permisison and Visibility view mixins

alexandria/core/views.py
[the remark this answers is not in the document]

alexandria/settings/alexandria.py
```diff
@@ -190,6 +190,15 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
     "ALEXANDRIA_DAV_SCHEME",
     default=default("ms-word:ofe|u|http", "ms-word:ofe|u|https"),
 )
+ALEXANDRIA_MANABI_ALLOWED_MIMETYPES = env.list(
+    "ALEXANDRIA_MANABI_ALLOWED_MIMETYPES",
+    default=[
+        "application/msword",
+        "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
+        "application/vnd.ms-excel",
+        "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
```
anehx, line 199:
```suggestion
        "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
        "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
```
Let's not enable `.doc` and `.xls` files per default

alexandria/core/serializers.py
Yelinz: Raises 404 in `check_permission` when not enabled

alexandria/core/serializers.py
Yelinz: `check_permissions` and `check_object_permissions` in the WebDAV view are being used to call DGAP on `get` request.

anehx (approved):

anehx:
> I know I'm a bit late to the party, but what was the reason for this? Is it so we can manage permissions for WebDAV separately? Woulda been nice to have this explained in a commit message

Yeah, we displayed WebDAV links for documents that should not be editable. If we'd checked those permissions in the existing view, the performance would be quite awful which is why we extracted it in a separate view.

PR #557, opened 2024-05-29
chore(deps): upgrade django to 4.2 and python to 3.10

JohnL17:


Dockerfile
```diff
@@ -38,6 +38,7 @@ RUN pip install -U poetry
 
 ARG INSTALL_DEV_DEPENDENCIES=false
 COPY pyproject.toml poetry.lock $APP_HOME/
+RUN poetry lock
```
Yelinz, line 41: should be removed

Yelinz (changes requested):
also needs change in dependabot to not ignore django >=4

Dockerfile
JohnL17: Ah sorry I forgot to remove it. It's removed now.

Yelinz (approved):

JohnL17:
@Yelinz I did something wrong when I did the rebase. So I changed that and now the tests run through again. But it would be great if you could review the changes again.

Yelinz:
Looked good still

PR #564, opened 2024-06-04
fix(clamav)!: inline django-clamd to resolve version problems

Yelinz:
resolves the problem of django-clamd not having a release with deprecation fixes
uses [clamdpy](https://github.com/Viicos/clamdpy) as it is a newer clamav binding with the same api as python-clamd (which django-clamd uses)


alexandria/core/validations.py
```diff
@@ -1,13 +1,48 @@
...
+        scanner.ping()
+
+        # Server is up. This means that the file is too big.
+        log.warning(
+            "The file is too large for ClamD to scan it. Bytes Read {}".format(
+                file.tell()
+            )
+        )
+        raise ValidationError(
+            _("Malware scan could not completed. Please try again later."),
```
winged, line 36: Needs a `manage.py makemessages` run I think. Also, potentially reword a bit..

```suggestion
            _("Malware scan could not completed. Please notify the responsible administrator."),
```

(Rewording, as just trying again will not help in this case, and the ClamD config likely needs to change)

winged (changes requested):
LGTM, just a small suggestion

alexandria/core/validations.py
Yelinz: Applied

Yelinz: Looked at clamdpy API a bit more and it handles too large files and other errors itself already.
Therefore the try catch was removed

winged: Even better :-)

alexandria/settings/alexandria.py
```diff
@@ -169,8 +169,6 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
 ALEXANDRIA_ENABLE_CHECKSUM = env.bool("ALEXANDRIA_ENABLE_CHECKSUM", default=True)
 
 # Clamav service
-CLAMD_SOCKET = env.str("ALEXANDRIA_CLAMD_SOCKET", default="/var/run/clamav/clamd.ctl")
-CLAMD_USE_TCP = env.bool("ALEXANDRIA_CLAMD_USE_TCP", default=True)
 CLAMD_TCP_SOCKET = env.str("ALEXANDRIA_CLAMD_TCP_SOCKET", default=3310)
 CLAMD_TCP_ADDR = env.str("ALEXANDRIA_CLAMD_TCP_ADDR", default="localhost")
 CLAMD_ENABLED = env.bool("ALEXANDRIA_CLAMD_ENABLED", default=False)
```
winged, line 174: I think since the code is inlined now, we could avoid this unexpected name mismatch between env and settings..

```suggestion
ALEXANDRIA_CLAMD_TCP_SOCKET = env.str("ALEXANDRIA_CLAMD_TCP_SOCKET", default=3310)
ALEXANDRIA_CLAMD_TCP_ADDR = env.str("ALEXANDRIA_CLAMD_TCP_ADDR", default="localhost")
ALEXANDRIA_CLAMD_ENABLED = env.bool("ALEXANDRIA_CLAMD_ENABLED", default=False)
```

docker-compose.yml
```diff
@@ -80,6 +83,13 @@ services:
     ports:
       - "9998:9998"
   
+  clamav:
+    image: tiredofit/clamav:2.6.17
+    ports:
+      - "3310:3310"
+    environment:
+      - CONCURRENT_DATABASE_RELOAD=false
+      - ENABLE_DEFINITIONS_UPDATE=false
```
winged, line 92: I just hope nobody copies this to prod somewhere :-)

winged (changes requested):
Nice work! Looking good, just some minor nitpicks as usual ;-)

alexandria/settings/alexandria.py
Yelinz: Done

winged (approved):

PR #567, opened 2024-06-06
chore(deps): upgrade django-localized-fields to 6.8b4

Yelinz:
needed for better dependency version resolving. Relaxes version range of deprecations and adds django 4 support.


open-dynaMIX (approved):
Might be worth adding a comment explaining why we're using a beta version here. Maybe link to the [issue](https://github.com/SectorLabs/django-localized-fields/issues/95).

PR #571, opened 2024-06-10
fix(tika): use module path imports

Yelinz:
has some weirdness when using `import tika` instead of the modules directly, sometimes the import works other times it errors with a module not found through `tika.parser`

### DIFF
--- a/alexandria/core/models.py
+++ b/alexandria/core/models.py
@@ -6,7 +6,8 @@ from mimetypes import guess_extension
 from pathlib import Path
 from tempfile import NamedTemporaryFile
 
-import tika
+import tika.language
+import tika.parser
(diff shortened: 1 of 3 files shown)


winged:
To be honest I preferred it the other way. `parser` and `language` could conceptually be from somewhere entirely different; using `tika.parser` and `tika.language` makes it 100% clear what module/package you're referring to in the code.

To avoid the problem you had, you can use `import tika.parser` and `import tika.language` as needed. These submodules aren't automatically imported when you're importing the top-level `tika` module, so explicit submodule import is a good thing IMO

Yelinz:
Changed to your suggestion :+1:

winged (approved):
Awesome :+1:

PR #572, opened 2024-06-13
fix(file): improve content vector generation command

Yelinz:
prevents cancellation of command if one of the files is not found

### DIFF
--- a/alexandria/core/management/commands/generate_content_vectors.py
+++ b/alexandria/core/management/commands/generate_content_vectors.py
@@ -17,12 +17,27 @@ class Command(BaseCommand):
                 )
             )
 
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -120,6 +120,7 @@ omit = [
     "manage.py",
     "alexandria/wsgi.py",
     "generate_missing_thumbnails.py",
+    "generate_content_vectors.py",
```
winged, line 123: I'd prefer to have this tested, except the memory bit...

Yelinz: Added the negative test

winged (approved):
Nice, LGTM!

PR #582, opened 2024-06-26
feat(webdav): allow opening webdav links with custom URI schemes

anehx:


alexandria/settings/alexandria.py
```diff
@@ -184,17 +184,20 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
 ALEXANDRIA_MANABI_DAV_URL_PATH = env.str(
     "ALEXANDRIA_MANABI_DAV_URL_PATH", default="/dav"
 )
-ALEXANDRIA_MANABI_DAV_SCHEME = env.str(
```
Yelinz, line 187: Counts as breaking right?

alexandria/core/models.py
```diff
@@ -289,7 +289,13 @@ def get_webdav_url(self, username, group, host="localhost:8000"):
         payload = (username, group, str(self.pk))
         token = Token(key, path, payload=payload)
 
-        return f"{settings.ALEXANDRIA_MANABI_DAV_SCHEME}://{host}{settings.ALEXANDRIA_MANABI_DAV_URL_PATH}/{token.as_url()}"
+        handler = settings.ALEXANDRIA_MANABI_DAV_URI_SCHEMES.get(
+            self.mime_type, "ms-word:ofe|u|"
```
Yelinz, line 293: why not no handler for undefined mime types?

Yelinz (changes requested):
could avoid breaking change if you say deprecated setting overrides new setting

alexandria/settings/alexandria.py
anehx: Technically, yes. However, as this setting was completely useless in the past (it automatically does the correct thing - no one would configure HTTP instead of HTTPS in production) I'd suggest not doing a breaking release for this.

If we say it must be a breaking change then I'd rather declare it as such instead of doing a fallback as you suggested above as the previous setting made absolutely no sense (we allow excel files but then tell the browser to open them in Word).

@open-dynaMIX @winged @czosel Any inputs from your side regarding that matter?

alexandria/core/models.py
anehx: Fair enough, we could trigger an [`ImproperlyConfigured`](https://docs.djangoproject.com/en/5.0/ref/exceptions/#improperlyconfigured) exception if no handler was defined.

Yelinz: Yeah sounds like a good idea :+1: instead of silently returning nothing

anehx: Done

Yelinz (approved):
im fine without a breaking change

alexandria/settings/alexandria.py
anehx: @open-dynaMIX @winged @czosel If you don't have an opinion regarding this topic I'd appreciate a short comment so we can go ahead with this.

winged: Given that the old setting didn't work, I don't think this needs to considered as breaking. Green light from my side

winged (approved):

PR #591, opened 2024-07-09
fix(file): improve content vector generation

Yelinz:
Improves the generation management command to be more resilient and less error prone on long runs.

Adds handling when tika does not detect any content.

### DIFF
--- a/alexandria/core/management/commands/generate_content_vectors.py
+++ b/alexandria/core/management/commands/generate_content_vectors.py
@@ -17,14 +17,18 @@ class Command(BaseCommand):
                 )
             )
 
+        # disable additional features to focus on vectorization
+        settings.ALEXANDRIA_ENABLE_CHECKSUM = False
+        settings.ALEXANDRIA_ENABLE_THUMBNAIL_GENERATION = False
+
         failed_files = []
-        for file in tqdm(
-            File.objects.filter(variant="original", content_vector__isnull=True)
-        ):
+        query = File.objects.filter(variant="original", content_vector__isnull=True)
+        # iterate over files in batches to prevent memory exhaustion
+        for file in tqdm(query.iterator(50), "Generating vectors", query.count()):
             try:
                 file.save()  # this will call set_content_vector
-            except FileNotFoundError as e:
-                failed_files.append(file.id)
+            except Exception as e:  # noqa: B902
+                failed_files.append(str(file.id))
                 self.stdout.write(
                     self.style.WARNING(f"Error processing {file.id}: {e}")
                 )
--- a/alexandria/core/models.py
+++ b/alexandria/core/models.py
@@ -240,19 +240,26 @@ class File(UUIDModel):
         return f"sha256:{hashlib.sha256(bytes_).hexdigest()}"
 
     def set_checksum(self):
+        if not settings.ALEXANDRIA_ENABLE_CHECKSUM:
+            return
+
         self.content.file.file.seek(0)
         self.checksum = self.make_checksum(self.content.file.file.read())
 
     def set_content_vector(self):
         self.content.file.file.seek(0)
         parsed_content = tika.parser.from_buffer(self.content.file.file)
+
+        name_vector = SearchVector(Value(Path(self.name).stem), weight="A")
+        if not parsed_content["content"]:
+            self.content_vector = name_vector
+            return
+
         # use part of content for language detection, beacause metadata is not reliable
         detected_language = tika.language.from_buffer(parsed_content["content"][:1000])
         self.language = detected_language
         config = settings.ISO_639_TO_PSQL_SEARCH_CONFIG.get(detected_language, "simple")
-        self.content_vector = SearchVector(
-            Value(Path(self.name).stem), config=config, weight="A"
-        ) + SearchVector(
+        self.content_vector = name_vector + SearchVector(
             Value(parsed_content["content"]),
             config=config,
             weight="B",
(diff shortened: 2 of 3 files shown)


alexandria/core/models.py
```diff
@@ -246,6 +246,11 @@ def set_checksum(self):
     def set_content_vector(self):
         self.content.file.file.seek(0)
         parsed_content = tika.parser.from_buffer(self.content.file.file)
+
+        if not parsed_content["content"]:
+            self.content_vector = SearchVector(Value(""))
```
Yelinz, line 251: Should a file with no content be searchable (through name)?

alexandria/core/management/commands/generate_content_vectors.py
```diff
@@ -17,13 +17,19 @@ def handle(self, *args, **options):
...
+        settings.ALEXANDRIA_ENABLE_CHECKSUM = False
+        settings.ALEXANDRIA_ENABLE_THUMBNAIL_GENERATION = False
+
         failed_files = []
-        for file in tqdm(
-            File.objects.filter(variant="original", content_vector__isnull=True)
-        ):
+        query = File.objects.filter(variant="original", content_vector__isnull=True)
+        # iterate over files in batches to prevent memory exhaustion
+        for file in tqdm(query.iterator(50), "Generating vectors", query.count()):
```
Yelinz, line 29: Bottleneck seems to be tika, tried implementing this with multithreading but it always timed out when parsing the content with tika.
Speed is about 2-3 files per second.

alexandria/core/models.py
winged: Only if the name is part of the searchable space in the "success case" as well. Otherwise, I'd leave it to the filters to find files/documents by name

alexandria/core/models.py
```diff
@@ -292,9 +301,6 @@ def get_webdav_url(self, username, group, host="localhost:8000"):
         return f"{settings.ALEXANDRIA_MANABI_DAV_SCHEME}://{host}{settings.ALEXANDRIA_MANABI_DAV_URL_PATH}/{token.as_url()}"
 
     def create_thumbnail(self):
-        if not settings.ALEXANDRIA_ENABLE_THUMBNAIL_GENERATION:
-            return
-
```
winged, line 297: I'd prefer to keep this here. Otherwise, all the call sites of this method will need to check the setting

alexandria/core/management/commands/generate_content_vectors.py
```diff
@@ -17,14 +17,20 @@ def handle(self, *args, **options):
                 )
             )
 
+        # disable additional features to focus on vectorization
+        checksum = settings.ALEXANDRIA_ENABLE_CHECKSUM
+        thumbnail = settings.ALEXANDRIA_ENABLE_THUMBNAIL_GENERATION
```
winged, line 22: I don't think you need to keep these, as it's changed in the management command's state only. No need to change it back later

alexandria/core/management/commands/generate_content_vectors.py
Yelinz: Might not even need the memory check anymore, ran without memory problems even after 1500 files.

alexandria/core/models.py
winged: I even think that we should do the full check (`self.variant == File.Variant.ORIGINAL and self.renderings.count() == 0`) in here as well. Make the call site need as little logic as possible

winged (changes requested):
Nice work catching those issues. I have some cleanup suggestions though

alexandria/core/models.py
Yelinz: Moved into methods

alexandria/core/management/commands/generate_content_vectors.py
Yelinz: Removed

alexandria/core/models.py
Yelinz: The name is also added with more weight to the searchables in the normal case.
Added the name to the content_vector even when content is None

alexandria/core/models.py
```diff
@@ -240,19 +240,32 @@ def make_checksum(bytes_: bytes) -> str:
...
+            return
+
         self.content.file.file.seek(0)
         self.checksum = self.make_checksum(self.content.file.file.read())
 
     def set_content_vector(self):
+        if not (
+            self.variant == File.Variant.ORIGINAL
+            and settings.ALEXANDRIA_ENABLE_CONTENT_SEARCH
+        ):
```
winged, line 253: [De Morgan's rule](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) might make it more readable/understandable

```suggestion
        if (
            self.variant != File.Variant.ORIGINAL
            or not settings.ALEXANDRIA_ENABLE_CONTENT_SEARCH
        ):
```

alexandria/core/models.py
```diff
@@ -292,7 +299,11 @@ def get_webdav_url(self, username, group, host="localhost:8000"):
         return f"{settings.ALEXANDRIA_MANABI_DAV_SCHEME}://{host}{settings.ALEXANDRIA_MANABI_DAV_URL_PATH}/{token.as_url()}"
 
     def create_thumbnail(self):
-        if not settings.ALEXANDRIA_ENABLE_THUMBNAIL_GENERATION:
+        if not (
+            self.variant == File.Variant.ORIGINAL
+            and self.renderings.count() == 0
+            and settings.ALEXANDRIA_ENABLE_THUMBNAIL_GENERATION
+        ):
```
winged, line 306: Again, De Morgan for improved human parsing

winged (approved):
Awesome work! I have a small suggestion to reduce brain workload when reading the code, but feel free to merge nonetheless

PR #597, opened 2024-07-17
feat(file): search over all configured stemmings

Yelinz:
This prevents files from being not findable when there is no content or a wrong language is detected.

This might make the search a little slower, as were not filtering for the language anymore. But files can always be found, even if the language which tika detected it not in the `LANGUAGE` setting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,10 @@ test: ## Test the backend
 lint: ## Lint the backend
 	@docker compose run --rm alexandria sh -c "poetry run black --check . && poetry run flake8"
 
+.PHONY: format
+format: ## format the backend
+	@docker compose run --rm alexandria sh -c "poetry run black . && poetry run isort ."
+
 .PHONY: bash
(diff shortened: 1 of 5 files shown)


winged:
Change LGTM, but what was the reason again to do this in the Viewset and not in the filterset?

Yelinz:
> Change LGTM, but what was the reason again to do this in the Viewset and not in the filterset?

i dont think there was a specific reason, should I move the logic?

winged:
> > Change LGTM, but what was the reason again to do this in the Viewset and not in the filterset?
> 
> i dont think there was a specific reason, should I move the logic?

I think it'd be cleaner. But let's keep it here and do a clean refactor PR later, to keep the diff reasonable

winged (approved):
Nice work, LGTM! We'll need to keep an eye on performance though, filtering across all languages may be problematic as it will multiply the workload for the DB

PR #618, opened 2024-08-20
feat: add celery to allow indexing in background

czosel:
Prevents timeouts when Tika needs to index very large files, which can take a long time.

Testing is done by mocking away the celery tasks. No system test with the broker is being done.


compose.override.yaml
```diff
@@ -47,5 +47,29 @@ services:
...
+    ports:
+      - "6900-7000:6900-7000"
+      - "8001:8000"
+    environment:
+      - CELERY_RDB_HOST=0.0.0.0. # to be able to telnet from outside the container
+      - PYTHONUNBUFFERED=1       # any non-empty string will do
+      - ENV=dev
+      - DEBUG=true
+      - ALEXANDRIA_DEV_AUTH_BACKEND=true
+      - ALEXANDRIA_ALLOW_ANONYMOUS_WRITE=true
```
anehx, line 65: This should be inherited from `alexandria-dev`

compose.yaml
```diff
@@ -88,7 +88,24 @@ services:
     ports:
       - "3310:3310"
 
+  redis:
+    image: bitnami/redis:7.4
```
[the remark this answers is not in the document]
Yelinz: changed to official

alexandria/core/models.py
```diff
@@ -268,12 +269,18 @@ def set_content_vector(self):
...
         self, force_insert=False, force_update=False, using=None, update_fields=None
     ):
         self.set_checksum()
-        self.set_content_vector()
+        if (
+            settings.ALEXANDRIA_ENABLE_CONTENT_SEARCH
+            and self.variant == File.Variant.ORIGINAL
+            and not self.content_vector
+        ):
+            set_content_vector.delay_on_commit(self.pk)
```
winged, line 283: I'd move the logic into a `post_save` signal handler to trigger the appropriate Celery task(s)

compose.override.yaml
```diff
@@ -47,5 +49,22 @@ services:
...
+    <<: *alexandria-dev
+    ports:
+      - "6900-7000:6900-7000"
+      - "8001:8000"
+    command:
+      [
+        "/bin/sh",
+        "-c",
+        "wait-for-it db:5432 -- wait-for-it redis:6379 -- ./manage.py migrate && watchmedo auto-restart -d . --recursive -p '*.py' -- celery -A alexandria worker -l INFO -E",
+      ]
```
winged, line 66: Time for an entrypoint script IMO

alexandria/core/tasks.py
```diff
@@ -0,0 +1,9 @@
+from celery import shared_task
+
+
+@shared_task
+def set_content_vector(file_pk):
+    from .models import File
+
+    file = File.objects.get(pk=file_pk)
+    file.set_content_vector()
```
winged, line 9: Why not move the entire logic here, out of the model? I think it'd be cleaner that way

alexandria/core/models.py
```diff
@@ -268,12 +269,18 @@ def set_content_vector(self):
             config=config,
             weight="B",
         )
+        self.save()
 
     def save(
         self, force_insert=False, force_update=False, using=None, update_fields=None
     ):
         self.set_checksum()
```
winged, line 277: Suggest moving the `set_checksum()` code into a Celery task as well. That way, it can (theoretically) run in parallel to the content vector calculation

winged (changes requested):
Nice work! I do have some suggestions regarding code organisation though

alexandria/core/models.py
Yelinz: Done

compose.override.yaml
Yelinz: Done

alexandria/core/tasks.py
Yelinz: Done

alexandria/core/models.py
Yelinz: Done

alexandria/core/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+    config = settings.ALEXANDRIA_ISO_639_TO_PSQL_SEARCH_CONFIG.get(
+        file.language, "simple"
+    )
+    file.content_vector = name_vector + SearchVector(
+        Value(parsed_content["content"].strip()),
+        config=config,
+        weight="B",
+    )
+
+    file.save()
```
winged, line 43: Something just popped into my mind: We're saving the whole file object here, potentially overwriting the checksum that may now run in parallel (and probably finished earlier).

Either we'll need to wrap the task into `@transaction.atomic` or explicitly update only the `content_vector` field. I'd prefer the latter:

```suggestion
    # Update only content vector, to avoid race with the checksum task
    File.objects.filter(pk=file.pk).update(content_vector=file.content_vector)
```

alexandria/core/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+    # different values should be set in tika as well
+    # https://github.com/CogStack/tika-service/blob/master/README.md#tika-parsers-configuration
+    parsed_content = tika.parser.from_buffer(
+        file.content.file.file, requestOptions={"timeout": 300}
+    )
+
+    name_vector = SearchVector(Value(Path(file.name).stem), weight="A")
+    if not parsed_content["content"]:
+        file.content_vector = name_vector
+        file.save()
```
winged, line 29:
```suggestion
        # Update only content vector, to avoid race with the checksum task
        File.objects.filter(pk=file.pk).update(content_vector=file.content_vector)
```

alexandria/core/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+
+    file.save()
+
+
+@shared_task
+def set_checksum(file_pk: str):
+    file = File.objects.get(pk=file_pk)
+    file.content.file.file.seek(0)
+    file.checksum = make_checksum(file.content.file.file.read())
+    file.save()
```
winged, line 51:
```suggestion
    # Update only content vector, to avoid race with the checksum task
    File.objects.filter(pk=file.pk).update(content_vector=file.content_vector)
```

winged (changes requested):
Code starts to look good, but I've discovered a potential issue with a race condition between the tasks. Code suggestion available

compose.yaml
```diff
@@ -88,7 +88,21 @@ services:
...
+    volumes:
+      - redis_data:/redis/data
+
+  celery:
+    <<: *alexandria
+    volumes:
+      - ./celery/entrypoint.sh:/entrypoint.sh
+    entrypoint: /entrypoint.sh
+    ports:
+      - "8001:8000"
```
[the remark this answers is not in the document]
Yelinz: Was due to reusing service definition. Changed to not need this.

alexandria/core/tasks.py
Yelinz: Makes sense, didnt consider the overwriting each other. Applied

Yelinz:
Odd behaviour with snapshot test of `test_api_list[TagViewSet]`, locally running all or alone has a different snapshot change. On CI the single run snapshot is expected.

winged (approved):

PR #625, opened 2024-08-29
fix!: make document title and description not localized

czosel:
BREAKING CHANGE: This turns the document's `title` and `description` into regular CharFields/TextFields, instead of localized ones, because localized fields do not make sense for user-generated data.

Run `./manage.py migrate_document_title_and_description <source_language>` after deployment to fix existing database records.


alexandria/core/management/commands/migrate_document_title_and_description.py
```diff
@@ -0,0 +1,22 @@
+from django.core.management.base import BaseCommand
+from pghstore import loads
```
anehx, line 2: Why do we need this additional dependency?

czosel: I need this in the migration to parse the existing history content.

pyproject.toml
```diff
@@ -39,6 +39,7 @@ requests = "^2.31.0"
 tika = "^2.6.0"
 tqdm = "^4.66.5"
 gunicorn = "^23.0.0"
+pghstore = "^2.0.2"
```
czosel, line 42:
```suggestion
# only needed in migrate_document_title_and_description.py, can be dropped in the future
pghstore = "^2.0.2"
```

anehx (approved):

PR #630, opened 2024-09-03
fix(convert): change file extension in document title on conversion

anehx:
This fixes a regression introduced in 39436b7739d8a06c9da2a51f7a7b5bed82a84d4c


czosel (approved):
:+1:

PR #631, opened 2024-09-03
feat: add presigning functions to public api

czosel:


czosel:
offline-approved by @StephanH90

PR #652, opened 2024-09-20
fix: mime type handling

czosel:
This does two things:
- Always deliver files with the mime type that was discovered during
  upload
- Enforce the three "signals" for mime types (Cotent-Type header, file
  extension, and actual file content) are consistent during file upload

### DIFF
--- a/alexandria/core/validations.py
+++ b/alexandria/core/validations.py
@@ -4,7 +4,7 @@ from mimetypes import guess_type
 import magic
 from clamdpy import ClamdNetworkSocket
 from django.conf import settings
-from django.utils.translation import gettext_lazy as _
+from django.utils.translation import gettext_lazy
 from generic_permissions.validation import validator_for
 from rest_framework.exceptions import ValidationError
 
@@ -27,10 +27,12 @@ def validate_file_infection(file_content):
     file_content.seek(0)
 
     if result.status == "FOUND":
-        raise ValidationError(_("File is infected with malware."), code="infected")
+        raise ValidationError(
+            gettext_lazy("File is infected with malware."), code="infected"
+        )
     elif result.status == "ERROR":
         raise ValidationError(
-            (_("Malware scan had an error: ") + result.reason),
+            (gettext_lazy("Malware scan had an error: ") + result.reason),
             code="incomplete",
         )
 
@@ -47,7 +49,7 @@ def validate_mime_type(mime_type, category):
         and mime_type not in category.allowed_mime_types
     ):
         raise ValidationError(
-            _(
+            gettext_lazy(
                 "File type %(mime_type)s is not allowed in category %(category)s."
                 % {"mime_type": mime_type, "category": category.pk}
             )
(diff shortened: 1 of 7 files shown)


alexandria/settings/alexandria.py
```diff
@@ -220,3 +220,11 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
 ALEXANDRIA_CONTENT_SEARCH_TYPE = env.str(
     "ALEXANDRIA_CONTENT_SEARCH_TYPE", default="phrase"
 )
+
+# Mime types that are considered safe for Content-Disposition: inline
+SAFE_FOR_INLINE_DISPONSITION = [
```
c0rydoras, line 225: this seems to be a typo

```suggestion
SAFE_FOR_INLINE_DISPOSITION = [
```

c0rydoras (changes requested):

alexandria/core/views.py
```diff
@@ -247,8 +247,12 @@ def download(self, request, pk=None):
             )
             obj = models.File.objects.get(pk=pk)
 
+            unsafe = obj.mime_type not in settings.SAFE_FOR_INLINE_DISPONSITION
```
c0rydoras, line 250: here as well

c0rydoras (changes requested):

c0rydoras:
otherwise this looks good :)

alexandria/settings/alexandria.py
czosel: Thanks, fixed!

alexandria/core/views.py
czosel: fixed

c0rydoras (approved):
looks good to me

winged (approved):
Nice work, LGTM!

alexandria/settings/alexandria.py
```diff
@@ -220,3 +220,11 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
...
     "ALEXANDRIA_CONTENT_SEARCH_TYPE", default="phrase"
 )
+
+# Mime types that are considered safe for Content-Disposition: inline
+SAFE_FOR_INLINE_DISPOSITION = [
+    "application/pdf",
+    "image/jpeg",
+    "image/png",
+    "image/gif",
+]
```
winged, line 230: Should probably be configurable via env var (but with the same, sane default)

alexandria/core/validations.py
```diff
@@ -62,17 +64,40 @@ def validate_file(self, data, context):
...
-
-        validate_mime_type(mime_type, data["document"].category)
-        data["mime_type"] = mime_type
+        content_type_header = data["content"].content_type
+        extension_type, _ = guess_type(data["name"])
+
+        if not content_type_header:  # pragma: no cover
+            raise ValidationError(gettext_lazy("Content-Type header is missing."))
+        if not extension_type:
+            raise ValidationError(gettext_lazy("The uploaded file has no extension."))
```
winged, line 73: Not exactly true: If I understand correctly, the meaning is that the file name extension could not be mapped to a known mime type

alexandria/core/validations.py
```diff
@@ -62,17 +64,40 @@ def validate_file(self, data, context):
...
+        if file_content_type != content_type_header:
+            raise ValidationError(
+                gettext_lazy(
+                    "Content type %(content_type)s does not match detected file content %(file_content_type)s."
+                    % {
+                        "content_type": content_type_header,
+                        "file_content_type": file_content_type,
+                    }
+                )
+            )
```
winged, line 98: I sort of like it, but also see some problems: If the upload for some reason mismatches the mime types, the user gets an error that they can't really do anything about.

winged (changes requested):
Nice work! I'm wondering what impact it has on end users though - do current browsers on all platforms send correct content-types for the common file formats? (I really don't know, it's not a rhetorical question meant to push back)

alexandria/core/validations.py
czosel: True, fixed the wording

winged (approved):

alexandria/core/validations.py
czosel: as discussed, we might have to "loosen" this up if it turns out to be too restrictive in practice.

PR #659, opened 2024-09-30
feat: filter files by category

czosel:


alexandria/core/filters.py
```diff
@@ -168,10 +168,21 @@ class FileFilterSet(FilterSet):
     document_metainfo = JSONValueFilter(field_name="document__metainfo")
     active_group = ActiveGroupFilter()
     files = BaseCSVFilter(field_name="pk", lookup_expr="in")
+    category = CategoriesFilter(field_name="document__category")
+    categories = CategoriesFilter(field_name="document__category")
```
winged, line 172: Do we really need both? They're functionally equal. I'd rather only have the plural if possible. (I notice the `DocumentFilterSet` also has both though, so maybe for consistency... :thinking: )

alexandria/core/tests/test_filters.py
```diff
@@ -385,3 +385,18 @@ def test_document_category_filters(
...
+
+
+def test_file_category_filter(db, admin_client, file_factory):
+    f1 = file_factory(name="1.pdf")
+    file_factory(name="2.pdf")
+
+    response = admin_client.get(
+        reverse("file-list"),
+        {"filter[categories]": f1.document.category_id, "filter[variant]": "original"},
+    )
```
winged, line 397: Need a test for `category` as well if we're adding it to the filterset

winged (changes requested):
Just a few minor issues, LGTM otherwise

PR #671, opened 2024-10-17
feat!: implement improved fulltext search

winged:
This makes the search endpoint resemble a different data type (not "files"
anymore!) Also, the search behaves now rather differently:

* Search results are ordered by search rank
* Search results get the search context (a fragment of the text)
  as part of the response
* File and document are referenced as related fields (and includes are available)

This should provide a highly performant, useful search that does what it should[tm]

BREAKING CHANGE: This changes the structure and type of the search
endpoint's data.

### Drive-By: chore: do not restart minio config container

When MC failed, this would restart the container forever, even when the Alexandria dev env was stopped

### Drive-By: feat(filters): add "only_newest" filter for files

This helps selecting files when we're searching (or otherwise looking
for files) and only want the newest version

### Drive-By:  feat(cmdline): search utility

This runs the search as it were run through the search endpoint.
Note this is mainly to be used for performance testing. No auth
support and no visibility support exists currently. If you enable
visibility / auth, it will likely break or not return anything.


alexandria/core/filters.py
```diff
@@ -168,45 +169,89 @@ class FileFilterSet(FilterSet):
...
 
-        latest_files = models.File.objects.filter(
-            document=OuterRef("document_id"), variant=models.File.Variant.ORIGINAL
-        ).order_by("-created_at")
-        queryset = queryset.filter(pk=Subquery(latest_files.values("pk")[:1])).filter(
-            search_query
+        queryset = queryset.annotate(
+            search_rank=SearchRank(F("content_vector"), search_query),
+            search_context=SearchHeadline(F("content_text"), search_query),
         )
```
winged, line 250: Note: It's possible that we'll need to do a `.filter(content_vector=search_query)` as well to improve performance (instead of ordering by `search_rank`, which "only" sorts the best matches first)

alexandria/core/filters.py
```diff
@@ -168,45 +169,89 @@ class FileFilterSet(FilterSet):
...
+            "only_newest",
+        ]
 
 
-class FileSearchFilterSet(FileFilterSet):
+class SearchFilterSet(FileFilterSet):
     query = CharFilter(method="search_files")
 
+    # Cut-off value, we don't want all the garbage matches
+    min_rank = NumberFilter(field_name="search_rank", lookup_expr="gte")
```
Yelinz, line 210: This does not seem to be used anywhere. Should maybe also have a default value?

alexandria/core/tests/test_search.py
```diff
@@ -0,0 +1,80 @@
...
+@pytest.mark.parametrize(
+    "filters,expected_name,num_queries",
+    [
+        (
+            {"filter[query]": "important", "filter[min_rank]": 0.1},
+            {"Bern.jpeg", "Paris.png"},
+            1,
+        ),
+        ({"filter[query]": "London", "filter[min_rank]": 0.1}, {"London.png"}, 1),
+        #        ({"filter[query]": "", "filter[min_rank]": 0.1}, set(), 0),
```
Yelinz, line 42: To be removed?

winged: gone now

alexandria/core/filters.py
winged: It was an idea to have a (configurable) cut-off in the frontend. I guess it's likely not going to be used tho

alexandria/core/filters.py
winged: Updated the code, we're now doing a `.filter()` call

winged:
Status update: I let some custom file/document factories run overnight, created over 400k files and then used the FTS in this branch to search.

Using the new commandline utility, I was able to run some tests - general query duration is between 0.05 and 0.2 seconds.

alexandria/core/filters.py
Yelinz: If its not being used, then it can be removed right

winged: You are absolutely right. Removed (incl tests)

alexandria/core/filters.py
```diff
@@ -168,50 +168,91 @@ class FileFilterSet(FilterSet):
...
 
-        latest_files = models.File.objects.filter(
-            document=OuterRef("document_id"), variant=models.File.Variant.ORIGINAL
-        ).order_by("-created_at")
-        queryset = queryset.filter(pk=Subquery(latest_files.values("pk")[:1])).filter(
-            search_query
-        )
+        queryset = queryset.annotate(
+            search_rank=SearchRank(F("content_vector"), search_query),
+            search_context=SearchHeadline(F("content_text"), search_query),
```
Yelinz, line 245: If the search matches with the file name but not content, then there will be no context. But that should be ok.

PR #691, opened 2024-11-12
fix(document): revalidate mimetypes after moving document to different category

velrest:


winged:
Python is not JS: `hasattr()` checks for an attribute, not dict key presence.

```python
foo={"a":1, "b":2}

print("a" in foo) # -> True
print("c" in foo) # -> False
print(hasattr(foo, 'a')) # -> False
print(hasattr(foo, 'keys')) # -> True
```

velrest:
> Python is not JS: `hasattr()` checks for an attribute, not dict key presence.
> 
> ```python
> foo={"a":1, "b":2}
> 
> print("a" in foo) # -> True
> print("c" in foo) # -> False
> print(hasattr(foo, 'a')) # -> False
> print(hasattr(foo, 'keys')) # -> True
> ```

Yeh, I already wrote with Yelin about this :sweat: . Out of curiosity, whats a nice way to check if the variable you have is a dict or a class object?

alexandria/core/tests/test_validation.py
```diff
@@ -21,17 +21,12 @@ def validate(self, data, context):
...
     # See if the validation had some effect
-    document.refresh_from_db()
-    assert document.created_by_group == "foobar"
-
-    # second, ensure validator is not called for File
-    url = reverse("file-detail", args=[file.pk])
-    admin_client.patch(url, json={})
-    assert call_counter["validate"] == 1
+    tag.refresh_from_db()
+    assert tag.created_by_group == "foobar"
```
Yelinz, line 32: What was the need for this change?

alexandria/core/validations.py
```diff
@@ -101,3 +101,17 @@ def validate_file(self, data, context):
         data["mime_type"] = content_type_header
 
         return data
+
+    @validator_for(Document)
+    def validate_document(self, data, context):
+        if context["request"].method == "PATCH" and "category" in data:
+            category = data["category"]
+            document = context["view"].get_object()
+            if hasattr(document, "category") and (
```
Yelinz, line 110: `hasattr` check should be unnecessary as a document should always have a category or you could just check `document.category is not None`

Yelinz:
> whats a nice way to check if the variable you have is a dict or a class object?

`C` is a defined class.
```python
>>> c = C()
>>> type(c)
<class '__main__.C'>
>>> d = {}
>>> type(d)
<class 'dict'>
```

then you can just check with `is`

```python
>>> type(d) is dict
True
>>> type(d) is C
False
```

alexandria/core/validations.py
velrest: I think i got some KeyErrors or something like this thats why i put this in but ill check again.

alexandria/core/tests/test_validation.py
velrest: Because the test was previously for `Document` but since we already have custom validations there this tests triggers both the validations from the test and the one i added.

alexandria/core/validations.py
velrest: I removed it

winged:
> [...]

Don't do direct type comparisons. Use `isinstance()` if you need to know whether something has a certain type:

```python
class Foo:
   pass
class Bar(Foo):
   pass

x = Bar()
y = Foo()

isinstance(x, Foo) # -> True
isinstance(x, Bar) # -> True
isinstance(y, Foo) # -> True
isinstance(y, Bar) # -> False

z = {}
isinstance(z, Foo) # -> False
isinstance(z, Bar) # -> False
isinstance(z, dict) # -> True
```

Yelinz:
right forgot `isinstance`

Yelinz (approved):

PR #694, opened 2024-11-15
fix(file): clone files with s3 copy

Yelinz:
Use boto3 built in copy for cloning s3 files instead of fetching and reuploading the same content.

This change should substantially speed up cloning large amount of documents in a real environment with network latency. Locally the difference is not noticeable through the docker network.


winged (approved):
Awesome!

PR #698, opened 2024-11-18
ci(ghcr): run ghcr on publish an not on created

velrest:
If a release was created from a draft release,
the created event is not fired but the published is.

### DIFF
--- a/.github/workflows/ghcr.yml
+++ b/.github/workflows/ghcr.yml
@@ -2,7 +2,7 @@ name: Container image
 
 on:
   release:
-    types: [created]
+    types: [published]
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ghcr.yml
```diff
@@ -2,7 +2,7 @@ name: Container image
 
 on:
   release:
-    types: [created]
+    types: [published]
```
Yelinz, line 5: the pypi workflow also needs this change

velrest: Done :ok_hand:

Yelinz (approved):

PR #706, opened 2024-11-25
fix(document): dont assume that a document has files

velrest:


alexandria/core/models.py
```diff
@@ -155,7 +155,8 @@ class Document(UUIDModel):
     date = models.DateField(blank=True, null=True)
 
     def get_latest_original(self):
-        return self.files.filter(variant=File.Variant.ORIGINAL).latest("created_at")
+        files = self.files.filter(variant=File.Variant.ORIGINAL)
+        return files.latest("created_at") if len(files) else None
```
Yelinz, line 159: There are a few more call sites which will need to check for `None` now

Yelinz (changes requested):

alexandria/core/models.py
velrest: I've updated the other calls. I'm now raising a `ValidationError` everywhere code assumes that files must be present. Not sure if thats the right way though.

velrest: Test coverage is failing, do you usually ignore Exceptions for coverage?

velrest: I've added `no cover` to the check for now.

velrest: Done

Yelinz:
Very nice work

PR #708, opened 2024-11-26
chore: release 7.0.0

velrest:


velrest:
Depends on #709

winged:
It'd probably make sense to rebase onto current master and include the newest changes / fixes from @Yelinz as well

PR #757, opened 2025-02-03
feat(copy): add api endpoint to copy a document

nlzet:
Adding a new API endpoint to copy a document with underlying files

### DIFF
--- a/alexandria/core/api.py
+++ b/alexandria/core/api.py
@@ -1,7 +1,10 @@
 import logging
+from os.path import splitext
 
 from django.conf import settings
 from django.core.files import File
+from django.db import transaction
+from django.utils.translation import gettext as _
 
 from alexandria.core import models
 from alexandria.core.validations import validate_file
@@ -9,6 +12,54 @@ from alexandria.core.validations import validate_file
 log = logging.getLogger(__name__)
 
 
+@transaction.atomic()
+def copy_document(
+    document: models.Document, user: str, group: str, category: models.Category
+):
+    """
+    Copy a document and all its original files to a new document.
+
+    This function eases the copying of documents by automatically setting important fields.
+    Uses `create_file` to copy the original document files.
+    """
+
+    basename, ext = splitext(document.title)
+    copy_suffix = _("(copy)")
+    document_title = f"{basename} {copy_suffix}{ext}"
+    new_document = models.Document.objects.create(
+        title=document_title,
+        description=document.description,
+        metainfo=document.metainfo,
+        category=category,
+        created_by_user=user,
+        created_by_group=group,
+        modified_by_user=user,
+        modified_by_group=group,
+    )
+
+    # Copying only the originals - create_file() will create the thumbnails
+    document_files = models.File.objects.filter(
+        document=document, variant=models.File.Variant.ORIGINAL.value
+    ).order_by("created_at")
+    new_files = []
+    for document_file in document_files:
+        new_files.append(
+            create_file(
+                name=document_file.name,
+                document=new_document,
+                content=document_file.content,
+                mime_type=document_file.mime_type,
+                size=document_file.size,
+                user=document_file.created_by_user,
+                group=document_file.created_by_group,
+                metainfo=document_file.metainfo,
+            )
+        )
+
+    return new_document
+
+
+@transaction.atomic()
 def create_document_file(
     user: str,
     group: str,
--- a/alexandria/core/serializers.py
+++ b/alexandria/core/serializers.py
@@ -5,7 +5,6 @@ from django.conf import settings
 from django.db.transaction import atomic
 from django.template.defaultfilters import slugify
 from django.utils import translation
-from django.utils.module_loading import import_string
 from generic_permissions.validation import ValidatorMixin
 from generic_permissions.visibilities import (
     VisibilityResourceRelatedField,
@@ -14,6 +13,8 @@ from generic_permissions.visibilities import (
 from rest_framework_json_api import serializers
 from rest_framework_json_api.relations import SerializerMethodResourceRelatedField
 from rest_framework_json_api.views import Serializer
+
+from alexandria.core.utils import get_user_and_group_from_request
 
 from . import models
 
@@ -27,15 +28,10 @@ class BaseSerializer(
 
     created_at = serializers.DateTimeField(read_only=True)
 
-    def get_user_and_group(self):
-        return import_string(settings.ALEXANDRIA_GET_USER_AND_GROUP_FUNCTION)(
-            self.context.get("request")
-        )
-
     def is_valid(self, *args, **kwargs):
         # Prime data so the validators are called (and default values filled
         # if client didn't pass them.)
-        user, group = self.get_user_and_group()
+        user, group = get_user_and_group_from_request(self.context.get("request"))
         self.initial_data.setdefault("created_by_group", group)
         self.initial_data.setdefault("modified_by_group", group)
         self.initial_data.setdefault("created_by_user", user)
@@ -45,7 +41,7 @@ class BaseSerializer(
     def validate(self, *args, **kwargs):
         validated_data = super().validate(*args, **kwargs)
 
-        user, group = self.get_user_and_group()
+        user, group = get_user_and_group_from_request(self.context.get("request"))
         validated_data["modified_by_user"] = user
         validated_data["modified_by_group"] = group
 
@@ -324,7 +320,7 @@ class WebDAVSerializer(BaseSerializer):
         request = self.context.get("request")
         host = request.get_host() if request else "localhost"
         scheme = request.scheme if request else "http"
-        user, group = self.get_user_and_group()
+        user, group = get_user_and_group_from_request(self.context.get("request"))
         return instance.get_latest_original().get_webdav_url(
             user, group, f"{scheme}://{host}"
         )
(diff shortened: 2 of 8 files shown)


anehx:
@winged Could you take a look at this?

alexandria/core/views.py
```diff
@@ -114,6 +116,35 @@ def update(self, request, *args, **kwargs):
...
+            category = get_object_or_404(models.Category, pk=category_id)
+        else:
+            category = document.category
+
+        copied_document, __ = copy_document(
+            document=document,
+            category=category,
+            user=user,
+            group=group,
+        )
```
winged, line 142: Could we maybe move this into a specialized serializer instead? This seems like a lot of manual work dealing with the request here. Or, at least, there should be `request.data` I think, which would help avoid dealing with JSON

winged (changes requested):
Nice work so far! I have some feedback, mostly minor things

alexandria/core/api.py
```diff
@@ -35,6 +36,55 @@ def verify_signed_components(
     )
 
 
+def copy_document(
+    document: models.Document, user: str, group: str, category: models.Category
+):
+    """
+    Copy a document and all its files to a new document.
```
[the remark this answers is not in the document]
nlzet: Changed as proposed

alexandria/core/api.py
```diff
@@ -35,6 +36,55 @@ def verify_signed_components(
...
+
+    This function eases the copying of documents by automatically setting important fields.
+    Uses `create_file` to copy the original document files.
+    """
+
+    document_title = (
+        f"{splitext(document.title)[0]} (copy){splitext(document.title)[1]}"
+    )
```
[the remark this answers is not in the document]
nlzet: Changed as proposed

alexandria/core/api.py
```diff
@@ -35,6 +36,55 @@ def verify_signed_components(
...
+        new_files.append(
+            create_file(
+                name=document_file.name,
+                document=new_document,
+                content=document_file.content,
+                mime_type=document_file.mime_type,
+                size=document_file.size,
+                user=document_file.created_by_user,
+                group=document_file.created_by_group,
+                **additional_file_attributes,
```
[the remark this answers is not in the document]
nlzet: True, this was not needed and can just be passed as proposed, changed now.

alexandria/core/views.py
```diff
@@ -114,6 +116,35 @@ def update(self, request, *args, **kwargs):
...
+    @action(
+        methods=["post"],
+        detail=True,
+        url_path="copy",
+    )
+    def copy(self, request, pk=None):
+        document = self.get_object()
+        user, group = import_string(settings.ALEXANDRIA_GET_USER_AND_GROUP_FUNCTION)(
+            request
+        )
```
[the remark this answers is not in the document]
nlzet: As proposed I created this method in a new `utils.py`, and replaced it everywhere where used (was also used in `serializers.py`, so also replaced in there).

alexandria/core/views.py
nlzet: I created a new specialized serializer to handle the request input, and convert the category parameter into a category object when provided. It now also uses `request.data` to not deal with the JSON anymore.

nlzet:
@winged pushed some changes and commented on the change requests

alexandria/core/tests/test_api.py
```diff
@@ -25,6 +26,89 @@ def test_create_document_file(db, category):
...
+    assert files[0].name == "Mee.pdf"
+    assert files[0].mime_type == "image/png"
+    assert files[0].size == 1
+    # files will retain the user/group of the original document
+    assert files[0].created_by_user == "Foo"
+    assert files[0].created_by_group == "Baz"
+
+    # new thumbnail for first file
+    assert files[1].document.pk == copied_doc.pk
+    assert files[1].variant == File.Variant.ORIGINAL
```
winged, line 94: If it's a thumbnail, why is it of variant `ORIGINAL`? I think the comments might be mis-ordered here.

alexandria/core/tests/test_views.py
```diff
@@ -398,6 +399,70 @@ def test_move_document_to_new_category(
...
+    to_category,
+    expected_status,
+):
+    category_not_allowed = category_factory.create(allowed_mime_types=["plain/text"])
+    category = category_factory()
+    document = document_factory(category=category)
+    file_factory.create(document=document, name="Image.jpeg", mime_type="image/jpeg")
+
+    if to_category == "non_existent":
+        target_category_pk = 999
```
winged, line 426: This may cause flaky test if for some reason a category with ID 999 will eventually exist.

For such things, it's better to call the `category_factory()`, take it's primary key, then delete it before commencing with the test. Alternatively, just `assert` that it's missing, with a comment explaining why

alexandria/core/tests/test_views.py
```diff
@@ -398,6 +399,70 @@ def test_move_document_to_new_category(
...
+        },
+    }
+
+    if to_category == "not_defined":
+        request_body["data"].pop("category")
+
+    url = reverse("document-copy", args=[document.pk])
+    response = admin_client.post(
+        url,
+        data=json.dumps(request_body),
```
winged, line 448: That shouldn't be necessary.

`admin_client.post(url, data=request_data)` should just work

alexandria/core/serializers.py
```diff
@@ -363,3 +359,23 @@ class Meta:
...
+            try:
+                category = models.Category.objects.get(pk=category_id)
+            except models.Category.DoesNotExist:
+                raise serializers.ValidationError("Category not found")
+        else:
+            category = None
+
+        validated_data["category"] = category
+
+        return validated_data
```
winged, line 381: Instead of having `data` as a `DictField`, I'd put the category as it's own `ResourceRelatedField`. Build a serializer for the whole request, not just for the category part.

winged (changes requested):

alexandria/core/tests/test_api.py
nlzet: Yeah I mixed up some ordering here, improved it now

alexandria/core/tests/test_views.py
nlzet: I like the approach of creating and deleting a category for the test, changed it now in this way

alexandria/core/tests/test_views.py
nlzet: Changed now

alexandria/core/serializers.py
nlzet: Took me some time to figure this one out and changed the request format a bit to work with the `ResourceRelatedField`. Can you verify if the new version looks good? If so I will update the request in the `ember-alexandria` MR (https://github.com/projectcaluma/ember-alexandria/pull/1079) as well to reflect this change.

winged: I wonder what we'd need to do to have the request be in a proper [JSONAPI](https://jsonapi.org/) format. Fine for me as it is now, but still I think I'd like it to be even more consistent :-)

winged:
Awesome, I like it! Two things I'd like to quickly discuss still - both not strictly neccessary IMO:

* The REST format of the request is not (yet) strictly JSONAPI. If possible, we should try to invest a little bit more time into that. (Maybe have @anehx have a look as well)
* The copy operation will re-generate the thumbnails, even though we already have them in the source document. I'd like to explore a way to avoid having them recalculated. I think it'd be a refactoring of the `create_file()` function.

But other than that, nice work :+1:

anehx:
> * The REST format of the request is not (yet) strictly JSONAPI. If possible, we should try to invest a little bit more time into that. (Maybe have @anehx have a look as well)

Makes sense, added a comment: https://github.com/projectcaluma/alexandria/pull/757#discussion_r1958161440

> * The copy operation will re-generate the thumbnails, even though we already have them in the source document. I'd like to explore a way to avoid having them recalculated. I think it'd be a refactoring of the `create_file()` function.

I agree that this would be the desired way. But I think we should keep it like this for now as it would otherwise be a larger refactoring.

alexandria/core/api.py
```diff
@@ -35,6 +37,53 @@ def verify_signed_components(
     )
 
 
+def copy_document(
```
anehx, line 40: It would make sense to define this as atomic using `@transaction.atomic`

alexandria/core/api.py
```diff
@@ -35,6 +37,53 @@ def verify_signed_components(
...
+                user=document_file.created_by_user,
+                group=document_file.created_by_group,
+                metainfo=document_file.metainfo,
+            )
+        )
+
+    return new_document
+
+
 def create_document_file(
```
anehx, line 87: If we're already on it: this should be atomic as well

anehx (approved):

alexandria/core/tests/test_views.py
```diff
@@ -398,6 +398,71 @@ def test_move_document_to_new_category(
...
+        }
+    }
+
+    if to_category == "not_defined":
+        data = {}
+    elif to_category == "null":
+        data["category"] = None
+
+    url = reverse("document-copy", args=[document.pk])
+    response = admin_client.post(url, data=data, format="json")
```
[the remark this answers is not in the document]
nlzet: Changed this format in the test, and also changed the request accordingly in https://github.com/projectcaluma/ember-alexandria/pull/1079

alexandria/core/api.py
nlzet: Changed

PR #780, opened 2025-02-27
fix(mark): fix data type and requiredness of mark name

anehx:
For the name property, we should use `LocalizedCharField` instead of `LocalizedTextField`. Also, the name must be optional since otherwise our migration needs to assume the languages of the host app which is simply not possible. This issue led to "missing migrations" warnings in consuming applications.


winged (approved):
Nice, LGTM

PR #798, opened 2025-03-24
fix(storages): use ssec storage at correct times

Yelinz:
fix document clone storage usage

### DIFF
--- a/alexandria/core/models.py
+++ b/alexandria/core/models.py
@@ -1,14 +1,12 @@
 import re
 import uuid
 from pathlib import Path
-from tempfile import NamedTemporaryFile
 
 from django.conf import settings
 from django.contrib.postgres.fields import ArrayField
 from django.contrib.postgres.indexes import GinIndex
 from django.contrib.postgres.search import SearchVectorField
 from django.core.exceptions import ImproperlyConfigured, ObjectDoesNotExist
-from django.core.files import File as DjangoFile
 from django.core.validators import RegexValidator
 from django.db import models, transaction
 from django.dispatch import receiver
@@ -18,7 +16,6 @@ from localized_fields.fields import LocalizedCharField, LocalizedTextField
 from manabi.token import Key, Token
 from rest_framework_json_api.relations import reverse
 
-from alexandria.storages.backends.s3 import S3Storage
 from alexandria.storages.fields import DynamicStorageFileField
 
 
@@ -171,38 +168,11 @@ class Document(UUIDModel):
 
         self.pk = None
         self.save()
+        new_name = f"{self.pk}_{latest_original.name}"
 
-        storage = File.content.field.storage
         latest_original.pk = None
         latest_original.document = self
-        if isinstance(storage, S3Storage):
-            new_name = f"{self.pk}_{latest_original.name}"
-            copy_args = {
-                "CopySource": {
-                    "Bucket": settings.ALEXANDRIA_S3_BUCKET_NAME,
-                    "Key": latest_original.content.name,
-                },
-                # Destination settings
-                "Bucket": settings.ALEXANDRIA_S3_BUCKET_NAME,
-                "Key": new_name,
-            }
-
-            if settings.ALEXANDRIA_ENABLE_AT_REST_ENCRYPTION:
-                copy_args["CopySourceSSECustomerKey"] = storage.ssec_secret
-                copy_args["CopySourceSSECustomerAlgorithm"] = "AES256"
-                copy_args["SSECustomerKey"] = storage.ssec_secret
-                copy_args["SSECustomerAlgorithm"] = "AES256"
-
-            storage.bucket.meta.client.copy_object(**copy_args)
-            latest_original.content = new_name
-            latest_original.save()
-        else:
-            with NamedTemporaryFile() as tmp:
-                temp_file = Path(tmp.name)
-                with temp_file.open("w+b") as file:
-                    file.write(latest_original.content.file.file.read())
-                    latest_original.content = DjangoFile(file)
-                    latest_original.save()
+        latest_original.content.copy(new_name)
 
         return self
 
--- a/alexandria/storages/fields.py
+++ b/alexandria/storages/fields.py
@@ -1,4 +1,8 @@
+from pathlib import Path
+from tempfile import NamedTemporaryFile
+
 from django.conf import ImproperlyConfigured, settings
+from django.core.files import File as DjangoFile
 from django.core.files.storage import storages
 from django.db import models
 from django.db.models.fields.files import FieldFile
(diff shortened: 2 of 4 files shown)


alexandria/core/models.py
```diff
@@ -172,7 +173,10 @@ def clone(self):
         self.pk = None
         self.save()
 
-        storage = File.content.field.storage
+        storage_backend = settings.ALEXANDRIA_FILE_STORAGE
+        if settings.ALEXANDRIA_ENABLE_AT_REST_ENCRYPTION:
+            storage_backend = "alexandria.storages.backends.s3.SsecGlobalS3Storage"
+        storage = storages.create_storage({"BACKEND": storage_backend})
```
fugal-dy, line 179: is this necessary?

`File.content.field.storage` is not based on the setting? okay, that's cool. The actual question is: why does `DynamicStorageFieldFile` logic not apply here. Or: do we get to decide on the storage to use at this point or not? A `clone` method should not be concerned with storage details at all. so I guess we'll have to get more abstract here.

The `File.content.field.storage` should do the exact same thing based on the setting `ALEXANDRIA_ENABLE_AT_REST_ENCRYPTION`  and the `alexandria/storages/fields.py` in `DynamicStorageFieldFile`'s `__init__`.

alexandria/storages/fields.py
```diff
@@ -4,20 +4,17 @@
...
-        self.storage = storages.create_storage(
-            {"BACKEND": settings.ALEXANDRIA_FILE_STORAGE}
-        )
+        storage_backend = settings.ALEXANDRIA_FILE_STORAGE
         if settings.ALEXANDRIA_ENABLE_AT_REST_ENCRYPTION:
             from alexandria.core.models import File
 
             if instance.encryption_status == File.EncryptionStatus.SSEC_GLOBAL_KEY:
-                self.storage = SsecGlobalS3Storage()
+                storage_backend = "alexandria.storages.backends.s3.SsecGlobalS3Storage"
```
fugal-dy, line 16: that's cool

alexandria/storages/fields.py
```diff
@@ -28,7 +25,9 @@ class DynamicStorageFileField(models.FileField):
     # of Alexandria: https://github.com/projectcaluma/alexandria/issues/480
 
     def __init__(self, storage=None, **kwargs):
-        storage = storages.create_storage({"BACKEND": settings.ALEXANDRIA_FILE_STORAGE})
+        self.storage = storages.create_storage(
+            {"BACKEND": settings.ALEXANDRIA_FILE_STORAGE}
+        )
         super().__init__(storage=storage, **kwargs)
```
fugal-dy, line 31: that's a bit hard to grasp. `storage` is not handled here but `self.storage` is. aka in the `super()` sth sth statement the storage arg is passed on or always `None` or the default is `None` which we clearly don't want, right? How are conflicting values of `self.storage` and `super().__init__(storage=storage)` behave

fugal-dy (changes requested):

alexandria/core/models.py
Yelinz: Cloning was done this way on purpose. This is to avoid downloading and reuploading the same file content. We want to be able to use the S3 api in the storage class.
See https://github.com/projectcaluma/alexandria/pull/694

alexandria/storages/fields.py
Yelinz: This is a mistake, reverted

alexandria/core/models.py
fugal-dy: I'm aware of that and I like it. My comment was not very clear, tbh. What I'm actually concerned with is: this is quite a lot of logic packed in the `clone` method that is not `clone` specific but storage specific.

I think it could be much clearer and easier to read if we could see here something like:

```
def clone: 
  make a copy of the document
  add a copy of the latest original file
  done
```

To achieve this I would suggest to e. g. add a `copy` method to the class `DynamicStorageFieldFile`

```python
class DynamicStorageFieldFile(FieldFile):
    def __init__(self, instance, field, name):
        super().__init__(instance, field, name)
        self.storage = storages.create_storage(
            {"BACKEND": settings.ALEXANDRIA_FILE_STORAGE}
        )
        storage_backend = settings.ALEXANDRIA_FILE_STORAGE
        if settings.ALEXANDRIA_ENABLE_AT_REST_ENCRYPTION:
            from alexandria.core.models import File

            if instance.encryption_status == File.EncryptionStatus.SSEC_GLOBAL_KEY:
                storage_backend = "alexandria.storages.backends.s3.SsecGlobalS3Storage"
        self.storage = storages.create_storage({"BACKEND": storage_backend})

    def copy(self, target_name):
        # S3 compatible storage: copy the file in storage without downloading and reuploading
        if isinstance(self.storage, S3Storage):
            copy_args = {
                "CopySource": {
                    "Bucket": self.storage.bucket,
                    "Key": self.name,
                },
                # Destination settings
                "Bucket": self.storage.bucket,
                "Key": target_name,
            } 
            if isinstance(self.storage, SsecGlobalS3Storage):
                copy_args["CopySourceSSECustomerKey"] = self.storage.ssec_secret
                copy_args["CopySourceSSECustomerAlgorithm"] = self.storage.customer_algorithm
                copy_args["SSECustomerKey"] = storage.ssec_secret
                copy_args["SSECustomerAlgorithm"] = self.storage.customer_alorithm
             
            self.storage.bucket.meta.client.copy_object(**copy_args)
            self.content = target_name
            self.save()
            return

        # otherwise use filesystem storage
        with NamedTemporaryFile() as tmp:
                temp_file = Path(tmp.name)
                with temp_file.open("w+b") as file:
                    file.write(self.read())
                    self.content = DjangoFile(file)
                    self.save()
               return
```

fugal-dy (changes requested):

alexandria/core/models.py
```diff
@@ -172,7 +173,10 @@ def clone(self):
         self.pk = None
         self.save()
 
-        storage = File.content.field.storage
+        storage_backend = settings.ALEXANDRIA_FILE_STORAGE
+        if settings.ALEXANDRIA_ENABLE_AT_REST_ENCRYPTION:
+            storage_backend = "alexandria.storages.backends.s3.SsecGlobalS3Storage"
```
fugal-dy, line 178: From what I understand this is the only place in your changes where storage selection is changed compared to before, right? This may however cause failure because the `DynamicStorageFieldFile`  relies on the `Field.encryption_status` to determine which `storage` to use. The `clone` method should do that as well. If it doesn't and the `settings.ALEXANDRIA_ENABLE_AT_REST_ENCRYPTION=True` will make the copy procedure fail for a `File` instance with `encryption_status` empty and content unencrypted, do you agree or am I getting this wrong?

fugal-dy (changes requested):

alexandria/core/models.py
Yelinz: Yes makes sense, I moved it into the field class

alexandria/core/models.py
Yelinz: Yes, the previous implementation was completely wrong and with moving it into the field class it also respects the `encryption_status`

alexandria/core/models.py
```diff
@@ -171,38 +168,11 @@ def clone(self):
...
-            latest_original.content = new_name
-            latest_original.save()
-        else:
-            with NamedTemporaryFile() as tmp:
-                temp_file = Path(tmp.name)
-                with temp_file.open("w+b") as file:
-                    file.write(latest_original.content.file.file.read())
-                    latest_original.content = DjangoFile(file)
-                    latest_original.save()
+        latest_original.content.copy(new_name)
```
fugal-dy, line 175: second thought on saving the parent object within the field content: appears somewhat over the top and is probably quite unexpected behaviour

```suggestion
        latest_original.content.copy(new_name)
        latest_original.save()
```

alexandria/storages/fields.py
```diff
@@ -10,14 +14,44 @@
...
+            self.instance.save()
+            return
+
+        # otherwise use filesystem storage
+        with NamedTemporaryFile() as tmp:
+            temp_file = Path(tmp.name)
+            with temp_file.open("w+b") as file:
+                file.write(self.read())
+                self.instance.content = DjangoFile(file, target_name)
+                self.instance.save()
```
fugal-dy, line 54:
```suggestion
```

alexandria/storages/fields.py
```diff
@@ -10,14 +14,44 @@
...
+            }
+            if isinstance(self.storage, SsecGlobalS3Storage):
+                copy_args["CopySourceSSECustomerKey"] = self.storage.ssec_secret
+                copy_args["CopySourceSSECustomerAlgorithm"] = "AES256"
+                copy_args["SSECustomerKey"] = self.storage.ssec_secret
+                copy_args["SSECustomerAlgorithm"] = "AES256"
+
+            self.storage.bucket.meta.client.copy_object(**copy_args)
+            self.instance.content = target_name
+            self.instance.save()
```
fugal-dy, line 45:
```suggestion
```

alexandria/storages/fields.py
```diff
@@ -56,8 +90,9 @@ def pre_save(self, instance, add):
...
                 )
                 raise ImproperlyConfigured(msg)
-            storage = SsecGlobalS3Storage()
             if instance.encryption_status == File.EncryptionStatus.SSEC_GLOBAL_KEY:
-                self.storage = storage
+                self.storage = storages.create_storage(
+                    {"BACKEND": "alexandria.storages.backends.s3.SsecGlobalS3Storage"}
+                )
         _file = super().pre_save(instance, add)
         return _file
```
fugal-dy, line 98: the variable and the underscore thingy are both uneccessary

```suggestion
        return super().pre_save(instance, add)
```

fugal-dy (dismissed):

alexandria/core/models.py
Yelinz: If it does not get saved a copy of the file will exist which doesnt belong to anywhere. I feel like that is worse behavior

fugal-dy: If you save it it's the same the other way round where the origial doesn't belong anywhere. Looks like there is not really a solution to avoid loose copies if copy is just called randomly :)

So we should at least point that out in a comment/docstring

fugal-dy: So fine with me to keep the saving within the field context

alexandria/storages/fields.py
```diff
@@ -10,14 +14,44 @@
...
         if settings.ALEXANDRIA_ENABLE_AT_REST_ENCRYPTION:
             from alexandria.core.models import File
 
             if instance.encryption_status == File.EncryptionStatus.SSEC_GLOBAL_KEY:
-                self.storage = SsecGlobalS3Storage()
+                storage_backend = "alexandria.storages.backends.s3.SsecGlobalS3Storage"
+        self.storage = storages.create_storage({"BACKEND": storage_backend})
+
+    def copy(self, target_name: str):
+        # S3 compatible storage: copy in same bucket with s3 copy
```
fugal-dy, line 26:
```suggestion
        ``` Copy file content to target.
        
        NOTE: this will copy the content and this copy will be the new reference of the filefield. This means
        that there will be no pointer to the original content, unless you made a copy of the parent model 
        instance beforhands by e. g. setting `file_instance.pk` to `None` before calling
        `file_instance.content.copy`.
        ```
        # S3 compatible storage: copy in same bucket with s3 copy
```

alexandria/core/models.py
Yelinz: Added a docstring

fugal-dy (approved):

PR #799, opened 2025-03-26
fix(file): prevent validation exception for presigned url expiry

Yelinz:
Should we try to redirect to leave the user on the blank browser default 403 page?

### DIFF
--- a/alexandria/core/tests/test_views.py
+++ b/alexandria/core/tests/test_views.py
@@ -484,6 +484,16 @@ def test_download_file(admin_client, file, presigned, expected_status):
     assert result.status_code == expected_status
 
 
+def test_download_file_expired(admin_client, file, freezer):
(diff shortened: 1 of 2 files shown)


winged:
Would it be possible (content negotiation etc) to serve the user a small HTML page that explains the error? Ideally of course with translation and everything, and possibly a little bit of styling, so it doesn't look like fefe's blog

anehx:
@winged 

> Would it be possible (content negotiation etc) to serve the user a small HTML page that explains the error? Ideally of course with translation and everything, and possibly a little bit of styling, so it doesn't look like fefe's blog

I wouldn't do that as alexandria is not a standalone application that should provide any HTML messages. I checked how minio does it and they just return an XML response so our JSON response should be fine as well IMO.

anehx (approved):

PR #816, opened 2025-04-17
Fix file copying

martinwe-adfinis:
This is a follow-up on both #757 and #798:

 * Pass the bucket *name* to Django storages (not the bucket *object*).
 * Copy the `FieldFile` object when copying a document's file, to prevent both files referencing the same object in the storage.

### DIFF
--- a/alexandria/core/management/commands/fix_copied_file_references.py
+++ b/alexandria/core/management/commands/fix_copied_file_references.py
@@ -1,0 +1,108 @@
+from collections import defaultdict
+
+import botocore
+from django.core.management.base import BaseCommand
+from tqdm import tqdm
+
+from alexandria.core.models import File
+
+
+class Command(BaseCommand):
+    help = "Fixes copied file references by creating a new copy."
+
+    def add_arguments(self, parser):
+        parser.add_argument("--dry", dest="dry", action="store_true", default=False)
+        parser.add_argument("--since", dest="since", type=str, default=None)
+
+    def handle(self, *args, **options):
+        filename_counts = defaultdict(list)
+        file_list = File.objects.filter(variant="original")
+
+        for tmp_file in tqdm(file_list.order_by("created_at").iterator()):
+            filename = tmp_file.content.name
+            filename_counts[filename].append(tmp_file.pk)
+
+        fix_count = 0
+        for filename, pks in filename_counts.items():
+            # Skip files that only have one reference.
+            pks_len = len(pks)
+            if pks_len <= 1:
+                continue
+
+            duplicates_query = File.objects.filter(pk__in=pks).order_by("created_at")
+
+            # Filter copies by file creation timestamp if provided
+            if options["since"]:
+                duplicates_query = duplicates_query.filter(
+                    created_at__gt=options["since"]
+                )
+
+            original_duplicate, *duplicates = duplicates_query
+
+            ignored_output = ""
+            duplicates_len = len(duplicates)
+            if duplicates_len < (pks_len - 1):
+                ignored_len = pks_len - duplicates_len
+                ignored_output = f"/{pks_len} (ignored {ignored_len} files older than {options['since']})"
+
+            checksum = original_duplicate.checksum
+            self.stdout.write(
+                self.style.SUCCESS(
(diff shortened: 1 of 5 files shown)


martinwe-adfinis:
This is more of a draft, this still needs:

 * Tests
 * Setting the correct name in the storage (just using `content.name` doesn't properly replace whitespaces etc.)

More importantly, however: The `Document` model provides a `clone()` method that appears to already do most of what we're doing in the API's `copy_document()` method. Is there anything that speaks against reusing that here?

nlzet:
@anehx added tests for the changes and a new proposed management command to fix existing copied file references

winged:
Please also add a new test that explicitly does the following procedure:

* Create a document & file
* Make a copy
* Delete original document & file
* Verify copy is still there and readable

nlzet:
@winged added a new test to explicitly test for content readability of copied files, after deleting the original document

alexandria/core/management/commands/fix_copied_file_references.py
```diff
@@ -0,0 +1,90 @@
...
+    help = "Fixes copied file references by creating a new copy."
+
+    def add_arguments(self, parser):
+        parser.add_argument("--dry", dest="dry", action="store_true", default=False)
+
+    def handle(self, *args, **options):
+        filename_counts = {}
+        for tmp_file in tqdm(
+            File.objects.filter(variant="original").order_by("created_at")
+        ):
```
winged, line 18: This could be a huge number of files, and in this way, we're loading them all into memory at the same time.

Querysets can be turned into iterators that load the data batch-by-batch from DB, improving memory usage:

```suggestion
        for tmp_file in tqdm(
            File.objects.filter(variant="original").order_by("created_at").iterator()
        ):
```

alexandria/core/management/commands/fix_copied_file_references.py
```diff
@@ -0,0 +1,90 @@
...
+            filename = tmp_file.content.name
+            if filename in filename_counts:
+                filename_counts[filename].append(tmp_file.pk)
+            else:
+                filename_counts[filename] = [tmp_file.pk]
```
winged, line 23: If you use a `defaultdict` for this, you can get rid of the `if` check here, and just call `.append()` directly.
For example:
```python
from collections import defaultdict
foo = defaultdict(list)
foo["test"].append(1)
...
```

alexandria/core/management/commands/fix_copied_file_references.py
```diff
@@ -0,0 +1,90 @@
...
+                        )
+                    )
+                    continue
+
+                current_filename = duplicate.content.name
+                new_filename = f"{duplicate.pk}_{duplicate.name}"
+
+                # Only fix copied files, skip the original
+                if current_filename == new_filename:
+                    continue
```
winged, line 49: This seems a bit fragile, no? You already know that the original is the first one in the duplicates QS, so why not make use of that fact instead?

Maybe something along the lines of `original, *duplicates = File.objects.filter(...)`

alexandria/core/tests/test_api.py
```diff
@@ -99,10 +101,76 @@ def test_copy_document_api(db, category, category_factory, same_category):
...
+        "Baz2",
+        "Mee2.pdf",
+        SimpleUploadedFile(
+            name="test2.png",
+            content=FileData.png800,
+            content_type="png",
+        ),
+        "image/jpeg",
+        2,
+    )
```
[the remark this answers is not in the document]
nlzet: Changed

alexandria/core/management/commands/fix_copied_file_references.py
nlzet: Added iterator

alexandria/core/management/commands/fix_copied_file_references.py
nlzet: Changed to defaultdict

alexandria/core/management/commands/fix_copied_file_references.py
nlzet: Changed to exclude the original from the collection in the for loop

winged (approved):
LGTM now, nice work!

PR #825, opened 2025-04-29
fix(copy): ignore checksum if duplicate or original does not have one

nlzet:
Together with @martinwe-adfinis we noticed older files did not have a checksum. So for the fix command when either the original or the new file does not have a checksum we'll skip the sanity check.


martinwe-adfinis:
As I was also involved here, I'm going to leave the final review to @winged (or someone else). :slightly_smiling_face: 

I wonder if it might make sense to add a management script to calculate the missing checksums (via separate PR, that is), to avoid unexpected behaviour in the future. Any thoughts, @winged (and maybe @YelinZ, @czosel)?

Yelinz:
Yeah makes sense to make a command to add missing checksum

nlzet:
Added new command to generate missing checksums

alexandria/core/management/commands/generate_missing_checksums.py
```diff
@@ -0,0 +1,39 @@
...
+                file.save(update_fields=["checksum"])
+
+            fix_count += 1
+            self.stdout.write(
+                self.style.SUCCESS(
+                    f"Generated checksum for file {file.pk} with checksum {checksum}"
+                )
+            )
+
+        self.stdout.write("")
```
Yelinz, line 31: Is this line intentional for spacing?

nlzet: Yes this was only added for a little bit more clarity in the output

Yelinz (approved):

PR #838, opened 2025-05-21
fix(webdav): close db connections if unused or obsolete

winged:
Django calls a function to close unused/obsolete DB connections before and after each request. In WebDAV, we don't use Django infrastructure to deal with requests, and therefore that cleanup code is not being called.

This introduces a WSGIDav middleware to fix that problem.

However since our unit tests run in transactions, it's a bit uglier (and less tested) than I'd like it to be...


czosel (approved):
:100:

PR #943, opened 2025-11-11
feat(deps): update DGAP to v3

anehx:


winged (commented):
LGTM, just.. do we really support DGAP v1 all the way to v3?

winged (approved):

PR #1004, opened 2026-05-27
ci(dependabot): add 7 day cooldown

derrabauke:


anehx (commented):
This means we get security patches 7 days later, right? Wouldn't it make sense to add a lower `semver-patch-days` additionally?

derrabauke:
> This means we get security patches 7 days later, right? Wouldn't it make sense to add a lower `semver-patch-days` additionally?

I think that defeats the purpose. Most of the time supply-chain attacks are targeting patch level releases, as they get merged more easily and partly unattended.

Further I think the cooldown will have no realistic effect to our app being vulnerable for a released CVE which in turn gets patched. Second we potentially get security alerts for those kinds of vulnerabilities and can take measures besides dependabot to fix it in less than 7 days. :man_shrugging:

Third point: `docker` and `github-actions` do not support the `semver-patch-days` config.

Quite a interesting read on the cooldown topic: https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns

btw: 7 days is the config proposed by [zizmor](https://docs.zizmor.sh/audits/#dependabot-cooldown)

anehx (approved):
Makes sense, thanks for elaborating @derrabauke

PR #1015, opened 2026-05-28
chore: use official clamav image

derrabauke:
`tiredofit/clamav` got deleted. So let's use the official image, as it already has everything we need.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -77,9 +77,11 @@ services:
       - "9998:9998"
 
   clamav:
-    image: tiredofit/clamav:2.6.17
+    image: clamav/clamav@sha256:f7954ca7ca13f0ebc301bb8a8b492a88db793192c48763a0b9a2ffb64dcf4bee
     ports:
       - "3310:3310"
+    volumes:
+      - clamav_data:/var/lib/clamav
 
   redis:
     image: redis:7.4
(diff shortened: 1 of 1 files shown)


compose.yaml
```diff
@@ -22,7 +22,7 @@ services:
       # https://hub.docker.com/_/postgres
       # - POSTGRES_PASSWORD=
     volumes:
-      - dbdata:/var/lib/postgresql/data
```
winged, line 25: Unrelated

compose.yaml
```diff
@@ -77,9 +77,11 @@ services:
       - "9998:9998"
 
   clamav:
-    image: tiredofit/clamav:2.6.17
+    image: clamav/clamav:latest
```
winged: Don't do latest (supply chain attack risk). Pin to a specific version, or better yet to a hash

winged (changes requested):

compose.yaml
derrabauke: Extracted to https://github.com/projectcaluma/alexandria/pull/1016

compose.yaml
derrabauke: `done`

winged:
Have now verified that it works, and it does (Tried with random string and the EICAR test signature, got a proper negative and positive response). Integration works :+1:

winged (approved):
LGTM now, tested and it works

winged:
Probably needs a rebase once #1016 is fixed and merged

winged:
Superseded by #1018 due to MR deadlock

PR #1016, opened 2026-05-29
chore(compose): adapt psql data dir mount to psql v18+ convention

derrabauke:
Error after pq update:
```
Important Change: the PGDATA environment variable of the image was changed to be version specific in PostgreSQL 18 and above⁠. For 18 it is /var/lib/postgresql/18/docker. Later versions will replace 18 with their respective major version (e.g., /var/lib/postgresql/19/docker for PostgreSQL 19.x). The defined VOLUME was changed in 18 and above to /var/lib/postgresql. Mounts and volumes should be targeted at the updated location. This will allow users upgrading between PostgreSQL major releases to use the faster --link when running pg_upgrade and mounting /var/lib/postgresql.
```

See https://aronschueler.de/blog/2025/10/30/fixing-postgres-18-docker-compose-startup/

---

TL;DR: psql 18+ changed the env var for the data dir.

 Instead use `/var/lib/postgresql` for volume mount. So docker will create sub dirs for every major pq version like `18/docker/`.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -22,7 +22,7 @@ services:
       # https://hub.docker.com/_/postgres
       # - POSTGRES_PASSWORD=
     volumes:
-      - dbdata:/var/lib/postgresql/data
+      - dbdata18:/var/lib/postgresql
 
   alexandria:
     <<: *alexandria-common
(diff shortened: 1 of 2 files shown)


compose.yaml
winged: We should pin PostgreSQL here as well

winged (commented):
LGTM, but an (unrelated, but important) note that we should probably address here as well

winged:
Superseded by #1018 due to MR deadlock

PR #969, opened 2025-12-18
feat(models): use UUID7 for primary keys

anehx:

PR #1017, opened 2026-05-29
feat(db): use uuidv7 for better performance

winged:
UUIDv7 are sequential-ish, and so put much less load on the DB index, improving INSERT performance as well as lookup speeds


winged:
Dropping, there's already #969

PR #1022, opened 2026-06-23
feat: add validation for minimum acceptable PDF version

trowik:


alexandria/core/validations.py
```diff
@@ -58,6 +60,22 @@ def validate_mime_type(mime_type, category):
...
+    pdf_version = PdfReader(content).pdf_header
+    match = re.search(r"%PDF-(\d+\.\d+)", pdf_version, re.IGNORECASE)
+
+    if not match:
+        raise ValidationError(gettext_lazy(f"Invalid PDF header: {pdf_version}"))
+
+    version_number = match.group(1)
+    if version_number < min_version:
+        raise ValidationError(
+            gettext_lazy(f"PDF version {pdf_version} is not supported")
```
anehx:
```suggestion
            gettext_lazy(f"PDF version {pdf_version} is not allowed")
```

It is supported, the host app just choses to not allow it

alexandria/settings/alexandria.py
```diff
@@ -170,6 +170,9 @@ def default(default_dev=env.NOTSET, default_prod=env.NOTSET):
 # Checksums
 ALEXANDRIA_ENABLE_CHECKSUM = env.bool("ALEXANDRIA_ENABLE_CHECKSUM", default=True)
 
+# Minimum accepted PDF version
+ALEXANDRIA_MIN_PDF_VERSION = env.str("ALEXANDRIA_MIN_PDF_VERSION", default="")
```
anehx:
```suggestion
ALEXANDRIA_MIN_PDF_VERSION = env.float("ALEXANDRIA_MIN_PDF_VERSION", default=None)
```

alexandria/core/validations.py
```diff
@@ -72,6 +90,11 @@ def validate_file(self, data, context):
         if not extension_type:
             raise ValidationError(gettext_lazy("Unknown file extension"))
 
+        # validate PDF version if minimum version is defined
+        min_version = settings.ALEXANDRIA_MIN_PDF_VERSION
+        if extension_type == "application/pdf" and min_version:
```
anehx:
```suggestion
        if extension_type == "application/pdf" and min_version := settings.ALEXANDRIA_MIN_PDF_VERSION:
```

I personally quite like the walrus operator for stuff like this - if you don't no need to change it :)

alexandria/core/tests/test_validations.py
```diff
@@ -55,3 +59,74 @@ def test_validate_file_infection_infected(
...
+def test_validate_pdf_version(
+    db,
+    expected_error,
+    expected_status,
+    pdf_content,
+    admin_client,
+    settings,
+    document_factory,
+):
+    settings.ALEXANDRIA_MIN_PDF_VERSION = "1.6"
```
anehx:
```suggestion
    settings.ALEXANDRIA_MIN_PDF_VERSION = 1.6
```

alexandria/core/validations.py
```diff
@@ -58,6 +60,22 @@ def validate_mime_type(mime_type, category):
     return True
 
 
+def validate_pdf_version(content, min_version):
```
anehx: Let's add a docstring and types

alexandria/core/validations.py
```diff
@@ -58,6 +60,22 @@ def validate_mime_type(mime_type, category):
...
+
+    if not match:
+        raise ValidationError(gettext_lazy(f"Invalid PDF header: {pdf_version}"))
+
+    version_number = match.group(1)
+    if version_number < min_version:
+        raise ValidationError(
+            gettext_lazy(f"PDF version {pdf_version} is not supported")
+        )
+    return True
```
anehx: IMO there's no reason to return anything

anehx (changes requested):
Some small changes required, but otherwise LGTM 👍 Great work!

anehx (changes requested):
As discussed, only the translations for the other languages are missing now

trowik:
> As discussed, only the translations for the other languages are missing now

The issue with updating the translation files using`make makemessages` command was caused by `f-strings` inside `gettext_lazy` - had to refactor that

alexandria/core/validations.py
```diff
@@ -58,6 +61,34 @@ def validate_mime_type(mime_type, category):
...
+                "Invalid PDF header: %(pdf_version)s" % {"pdf_version": pdf_version}
+            )
+        )
+
+    version_number = float(match.group(1))
+    if version_number < min_version:
+        raise ValidationError(
+            gettext_lazy(
+                "PDF version %(pdf_version)s is not allowed"
+                % {"pdf_version": pdf_version}
```
anehx: Only thought about this now, sorry: Let's add the minimum version to the error message so the user is informed which versions work. Otherwise we just say "not allowed, try again" which is not very useful.

Maybe something like "PDF version 1.2 is not allowed, the minimum required PDF version is 1.6"

anehx (changes requested):

alexandria/core/validations.py
trowik: totally agree with that

alexandria/core/validations.py
```diff
@@ -58,6 +61,35 @@ def validate_mime_type(mime_type, category):
...
+            )
+        )
+
+    version_number = float(match.group(1))
+    if version_number < min_version:
+        raise ValidationError(
+            gettext_lazy(
+                "PDF version %(pdf_version)s is not allowed. "
+                "The minimum required PDF version is %(min_version)s"
+                % {"pdf_version": pdf_version, "min_version": min_version}
```
anehx:
```suggestion
                % {"pdf_version": version_number, "min_version": min_version}
```

anehx (approved):
🚀

trowik:
Had to rebase due to merge conflicts in the `poetry.lock` file

